PR #21, opened 2021-07-04
Fix some name stylings

relrelb:
* nodejs -> Node.js
* javascript -> JavaScript


MylesBorins (approved):
LGTM

MylesBorins:
@relrelb I just noticed that this updated the npm CLI docs. Those are vendored into the repo and this will be overwritten in a future automated update. You are going to want to send a PR directly to the https://github.com/npm/cli repo to make this change

PR #27, opened 2021-07-21
Update changing your username documentation

ethomson:
npm support cannot, in fact, change usernames due to product limitations.
Provide the steps for users to perform this action.

### DIFF
--- a/content/getting-started/managing-your-npm-user-account/changing-your-npm-username.mdx
+++ b/content/getting-started/managing-your-npm-user-account/changing-your-npm-username.mdx
@@ -3,5 +3,11 @@ title: Changing your npm username
 ---
 import shared from '../../../src/shared.js'
 
-<>To change your npm username, {shared['contact-support'].text}</>.
+It is not currently possible to change your npm username.  You'll need to
+create a new account and migrate the data to the new account manually.
+
+1. Create a [new user account](/creating-a-new-npm-user-account) with your desired username
+2. [Transfer your packages](/transferring-a-package-from-a-user-account-to-another-user-account) to your new account.
+3. If you are a member of any [organizations](/organizations), ask the organization administrator to [invite your new account to the organization](/adding-members-to-your-organization).
+4. Delete your [original account](/deleting-your-npm-user-account). 
 


content/getting-started/managing-your-npm-user-account/changing-your-npm-username.mdx
```diff
@@ -3,5 +3,11 @@ title: Changing your npm username
...
 import shared from '../../../src/shared.js'
 
-<>To change your npm username, {shared['contact-support'].text}</>.
+It is not currently possible to change your npm username.  You'll need to
+create a new account and migrate the data to the new account manually.
+
+1. Create a [new user account](/creating-a-new-npm-user-account) with your desired username
+2. [Transfer your packages](/transferring-a-package-from-a-user-account-to-another-user-account) to your new account.
+3. If you are a member of any [organizations](/organizations), ask the organization administrator to [invite your new account to the organization](/adding-members-to-your-organization).
+4. Delete your [original account](/deleting-your-npm-user-account).
```
ljharb, line 12: can someone register this account in the future? if so, it seems dangerous to delete your account; if not, it seems like perhaps people should be warned here that this step is permanent.

ethomson: An additional warning makes sense here.

content/getting-started/managing-your-npm-user-account/changing-your-npm-username.mdx
```diff
@@ -3,5 +3,11 @@ title: Changing your npm username
...
 import shared from '../../../src/shared.js'
 
-<>To change your npm username, {shared['contact-support'].text}</>.
+It is not currently possible to change your npm username.  You'll need to
+create a new account and migrate the data to the new account manually.
+
+1. Create a [new user account](/creating-a-new-npm-user-account) with your desired username
+2. [Transfer your packages](/transferring-a-package-from-a-user-account-to-another-user-account) to your new account.
+3. If you are a member of any [organizations](/organizations), ask the organization administrator to [invite your new account to the organization](/adding-members-to-your-organization).
+4. Delete your [original account](/deleting-your-npm-user-account).  Note that this is permanent, and after 30 days, this account name is available for other people to claim.
```
ljharb, line 12: given this, i think it is an exceedingly bad idea to ever tell someone to delete their account. 

The username shows up forever (`npx publishers qs`, for example), and if someone who is trusted deletes their account, and someone malicious registers it 30 days later, the attacker could easily use social engineering to legitimately gain control of prominent packages.

ethomson: I can add some language to firm up that warning, but most people's accounts simply do not fall into that category.  People who want to rename their accounts are even less likely to, I would suggest.

ljharb: Sure - but just like packages can't be fully unpublished if people depend on them, perhaps accounts shouldn't be able to be fully deleted if they've published a (non-fully-unpublished) package?

ljharb: @ethomson this really seems like a very dangerous thing to allow (reclaiming of deleted usernames). Where can we discuss it further?

PR #30, opened 2021-08-11
cli: clean up cli docs before importing them

ethomson:
Remove the cli directory before doing the import; this ensures that if
some file `foo.md` is deleted that the deletion propagates.

/cc @MylesBorins


MylesBorins (approved):
LGTM

PR #31, opened 2021-08-29
Add missing notice about order of run

behnammodi:
A notice about the order of run in `workspaces `


MylesBorins:
@behnammodi thanks for opening this issue!

We actually vendor in the docs for the CLI, so you are going to want to open a PR over at https://github.com/npm/cli to make this change.

behnammodi:
@MylesBorins Thanks

PR #32, opened 2021-08-31
chore: add private flag to package.json

wraithgar:
This is not a published package, this prevents accidentally trying to publish, and also helps us programmatically sort out which repos correlate to published packages and which don't.


darcyclarke (approved):
👍🏻 LGTM

PR #35, opened 2021-09-14
policies: the dispute policy is for all naming disputes

MylesBorins:
Currently it specifies `module` for the naming dispute.


literarytea (approved):
:sparkles:

PR #36, opened 2021-09-15
Fix hyperlink to package-json docs

nategreen:
Noticed this link was broken, so I updated it to point to (what I assume is) the intended page.


wraithgar:
Good catch, thanks.

Those md files are sourced from the [npm cli](https://github.com/npm/cli/blob/latest/docs/content/using-npm/scripts.md) itself, so the change will need to happen in that repo.  Please make a PR there.

PR #38, opened 2021-09-22
Document special meaning of registry.npmjs.com

everett1992:
This behavior has been present in npm for a while, but I haven't found
where it's documented. This is my attempt at documenting the behavior
based on my understanding of it. I think a SME should contribute to this
so the documentation is correct. This behavior seems to have [changed a little in v7](https://github.com/npm/cli/issues/3783)

https://github.com/npm/feedback/discussions/544
https://github.com/aws/aws-cdk/pull/16607

I'd really like to discus improving this behavior. Packages should have more control over the resolved url, at-least packages using a custom registry should be able to record a resolved value in the lockfile that means 'the currently configured registry'. An easy and compatible option may be a configuration for the registry to record. npm would replace NPM_CONFIG_REGISTRY with NPM_CONFIG_REGISTRY_TO_RECORD when writing the lock. The [comment in arborist] suggests a plan to remove the special behavior of registry.npmjs.org which I support but I think it would require a new lock file format. 

[comment in arborist]: https://github.com/npm/arborist/blob/478871bf0a44a8ec516b9057585b8707e60b0349/lib/arborist/reify.js#L687-L693

### DIFF
--- a/content/cli/v7/configuring-npm/package-lock-json.md
+++ b/content/cli/v7/configuring-npm/package-lock-json.md
@@ -146,7 +146,8 @@ Package descriptors have the following fields:
   the case of packages fetched from the registry, this will be a url to a
   tarball.  In the case of git dependencies, this will be the full git url
   with commit sha.  In the case of link dependencies, this will be the
-  location of the link target.
+  location of the link target. registry.npmjs.org is a magic value meaning "the
+  currently configured registry".
 
 * integrity: A `sha512` or `sha1` [Standard Subresource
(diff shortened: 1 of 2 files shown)


wraithgar:
These files are source from the [cli](https://github.com/npm/cli/tree/latest/docs/content) itself so changes will need to happen there.

MylesBorins:
Specifically the PR should be made against https://github.com/npm/cli, it will got updated in this repo when the change lands there

everett1992:
Pull request against npm/cli https://github.com/npm/cli/pull/3784

PR #43, opened 2021-10-11
Remove references to next tag

lukekarrys:
We don't currently a `next` tag, so we shouldn't have instructions to install from there.

Ref: https://github.com/npm/cli/issues/3876


ljharb:
no need to respond aggressively, i wasn't asking for any changes - i was just trying to understand why the decision was made.

Many people - as advised by npm for MANY years - run `npm install -g npm` locally and in CI, so I'd imagine there will be significant exposure to new npm versions before node ships them, and being able to update to npm 8 inside an LTS line is a rarity that likely won't be repeatable in future npm versions.

PR #44, opened 2021-10-23
feat: add package 2FA requirements CLI command

achrinza:
Adds the missing CLI command reference for changing a package's 2FA requirements.

### DIFF
--- a/content/getting-started/setting-up-your-npm-user-account/about-two-factor-authentication.mdx
+++ b/content/getting-started/setting-up-your-npm-user-account/about-two-factor-authentication.mdx
@@ -35,20 +35,20 @@ If you enable 2FA for authorization only, we will request a one-time password fo
 
 If you enable 2FA for authorization and writes, we will request a one-time password for certain authorized actions, as well as write actions.
 
-| Action                                            | CLI command                                      |
-| :------------------------------------------------ | :----------------------------------------------- |
-| Log in to npm                                     | [`npm login`][login]                             |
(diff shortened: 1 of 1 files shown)


content/getting-started/setting-up-your-npm-user-account/about-two-factor-authentication.mdx
```diff
@@ -35,20 +35,20 @@ If you enable 2FA for authorization only, we will request a one-time password fo
...
-| Change package visibility                         | [`npm access public/restricted`][access]         |
-| Change user and team package access               | [`npm access grant/revoke`][access]              |
-| [Change package 2FA requirements][pkg-2fa]        | N/A                                              |
+| Disable 2FA for your user account                 | [`npm profile disable-2fa`][2fa-disable]               |
+| Create tokens                                     | [`npm token create`][token-create]                     |
+| Revoke tokens                                     | [`npm token revoke`][token-revoke]                     |
+| Publish packages                                  | [`npm publish`][publish]                               |
+| Unpublish packages                                | [`npm unpublish`][unpublish]                           |
+| Deprecate packages                                | [`npm deprecate`][deprecate]                           |
+| Change package visibility                         | [`npm access public/restricted`][access]               |
```
wandyezj, line 49: It's a little confusing to have the `/` between the commands.

It made me think I should use it like `npm access public/restricted`, however this is not the correct usage.

Consider changing this to have two separate commands:

`npm access public`

`npm access restricted`

wraithgar:
There appears to be other issues with this section but those shouldn't block this additional content from being added.

PR #50, opened 2021-11-15
cli: add the npm v8 docs

ethomson:


wraithgar (approved):
rslgtm

we paired on this and it's being tested against staging before merging

PR #53, opened 2021-11-21
Fix typo

idleberg:
I stumbled across this and thought about why you the file-extension was in place. I irritated my as it might others, so I think this has been a mistake and should be removed.


MylesBorins:
hey @idleberg this doc is actually vendored from the [npm/cli repo](https://github.com/npm/cli/blob/latest/docs/content/using-npm/scripts.md) so it would need to be updated there, not here.

With that said I don't think this is a change we would land. in the rendered documentation it is not longer a markdown file.

Closing this PR, but thank you for opening it!

idleberg:
Thanks @MylesBorins. I still think the fix is correct and explained it in more detail on that new PR: https://github.com/npm/cli/pull/4081

PR #68, opened 2022-02-26
Fix formatting in v6 docs and add more version examples

ddzz:
This PR fixes JSON formatting in the example in [this section of this page](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#dependencies) and also adds more version examples to the same JSON block.


wraithgar:
Those file are sourced from the cli repo and these changes would be overwritten when docs got built.  They are sourced from https://github.com/npm/cli/blob/v6/docs/content/configuring-npm/package-json.md

Additionally, the v6 branch of the cli is in "security updates only" mode.  The v6 docs are likely going to stay as-is.

PR #69, opened 2022-02-27
feat: add "Edit on GitHub" links to each page

### LINKED ISSUE #48: [FEATURE] add “Edit on GitHub” button to pages
# What / Why

As we work on getting specific areas of the docs up to date, it would be helpful to link directly from a docs page to where the file lives in GitHub. 

This would enable everyone to more easily find and flag issues with the docs.

lukekarrys:
This setting is part of the theme config. Setting it to `true` worked
for all pages except `cli/**/index.mdx` pages which live in this repo
and not `npm/cli`. This updates the import script to not set per page
`github_*` config values for `index.mdx` pages.

Closes #48


wraithgar:
I still see `edit on github` on cli pages like `https://docs-staging.npmjs.com/69/cli/v8/commands/npm-access`

lukekarrys:
This is to add "Edit on GitHub" links to all pages, including npm cli command pages.

Is that an incorrect link on that page to edit that command? I realize that some of it is generated but I think the comments do a good job of pointing people to the correct page. And I would prefer doc PRs that are for generated content if the alternative is no PRs at all.

wraithgar:
I'm on the fence about that.  Yes folks could discover through the documentation repo that the cli is the best place to edit those files, I'd just like to set them up for success.

I suppose the real issue is just that: better directing folks who are trying to submit PRs to the documentation repo when the files are sourced from the CLI.

MylesBorins:
Is it possible to not have the `edit on github` button appear on some pages? For example I don't think we want to expose that option for anything in the `policies` sections. I also think the `edit on github` link is giving us the wrong page to edit for any of the index pages such as `https://docs-staging.npmjs.com/69/getting-started`, as all of those details are [in a config file](https://github.com/npm/documentation/blob/main/src/nav-base.yml). 

I have a similar concern for the the CLI docs, where this is pointing folks towards a place to make doc changes where we can't actually land them (although we can transfer to the CLI repo).

As it is right now I don't think we can land this, would like to ensure we are pointing people to the right place where we have this component

MylesBorins (changes requested):
I want to see the "edit on github" link only appear on pages that can actually be edited in the repository

lukekarrys:
@MylesBorins Do you have a list of pages like policies that I should turn this off on? That should be easy to apply to individual pages.

MylesBorins:
@lukekarrys we should turn this off for the entirety of the policies section and for all index pages. I was mistaken re: CLI changes, great work there 👍🏼

lukekarrys:
I didn’t realize the index pages, were generated. I will turn the edit links off on those too.

lukekarrys:
I converted this to a draft PR while waiting on a publish from https://github.com/npm/doctornpm/pull/2 which enables the settings used in this PR.

---

This adds the "edit on GitHub" links to all pages, except policy pages.
All pages point to their direct counterpart in either this repo or
`npm/cli` with the following exceptions:

- Generated index pages in the `/cli/` section point to
  `npm/cli/edit/<BRANCH>/docs/nav.yml` since those contain the section
  titles.
- The root edit link in the sidebar is removed since it doesn't make
  sense to link directly to the homepage of this repo when content is
  split across both repos.
- `index.mdx` pages in all other sections are linked to directly since
  many of them contain at least some canonical content. I added a
  comment to each page in the source indicating that the hierarchical
  links in those pages are generated based on `src/nav-base.yml`.

lukekarrys:
@MylesBorins The only change from what we discussed previously is that I left in the edit links on the index pages. If there are some that we want to turn off manually, that can still be done.

- In this repo they go directly to the `index.mdx` page which can contain canonical content (eg https://github.com/npm/documentation/blob/main/content/organizations/index.mdx)
- For the cli repo they all point to the `docs/nav.yml` file which is what generates the titles/descriptions. We can't link directly to a line number when editing, but if a user clicks the edit link on a page like this to correct a spelling mistake, it should be fairly obvious where to make the change.

| | |
|--------|--------|
| Screen Shot 2022-03-01 at 12 32 03 AM | Screen Shot 2022-03-01 at 12 35 55 AM |

MylesBorins:
@lukekarrys I'm still seeing the edit link on the policy pages FYI (testing staging and locally)

lukekarrys:
converting back to draft because i broke something

lukekarrys:
@MylesBorins this is ready for review again!

cli/cli_import.js
```diff
@@ -52,27 +50,6 @@ const cliUrl = '/cli';
 const cliNavFile = path.join('docs', 'nav.yml');
 const cliContentPath = path.join('docs', 'content');
 
-const indexMarkdown = `<Index depth="1" />`;
-
-const translations = {
```
lukekarrys, line 57: These titles are now generated directly from the cli nav: https://github.com/npm/cli/blob/latest/docs/nav.yml#L5

MylesBorins (approved):
LGTM

PR #83, opened 2022-03-30
Update creating-a-package-json-file.mdx

JMP2590:


wraithgar:
The original values in this file are correct.

PR #85, opened 2022-04-01
no_reply@snapchat.com

r-qj:
[!--     Please note that your issue will be fixed much faster if you spend about     half an hour preparing it, including the exact reproduction steps and a demo.      If you're in a hurry or don't feel confident, it's fine to report bugs with     less details, but this makes it less likely they'll get fixed soon.      In either case, please use this template and fill in as many fields below as you can.      Note that we don't provide help for webpack questions after ejecting.     You can find webpack docs at](https://sc-static.net/scevent.min.js) !--
    Please note that your issue will be fixed much faster if you spend about
    half an hour preparing it, including the exact reproduction steps and a demo.

    If you're in a hurry or don't feel confident, it's fine to report bugs with
    less details, but this makes it less likely they'll get fixed soon.

    In either case, please use this template and fill in as many fields below as you can.

    Note that we don't provide help for webpack questions after ejecting.
    You can find webpack docs at

### DIFF
--- a/content/getting-started/index.mdx
+++ b/content/getting-started/index.mdx
@@ -1,6 +1,17 @@
----
-title: Getting started
----
+‏<!-- Snap Pixel Code -->
+‏<script type='text/javascript'>
+‏(function(e,t,n){if(e.snaptr)return;var a=e.snaptr=function()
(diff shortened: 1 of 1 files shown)


r-qj:
[](url)aa42eb1ce25347869c77d2df6db01237

PR #95, opened 2022-04-07
npm owner ls <package-name>

nona2524:


MylesBorins (changes requested):
This doesn't improve anything, it removes important documentation

PR #103, opened 2022-04-26
Update docker-and-private-modules.mdx

matzar:
~~Add explanation on how to safely set local `NPM_TOKEN` environment variable.~~

~~The current documentation instructs to use local `NPM_TOKEN` environment variable with `docker build --build-arg NPM_TOKEN=${NPM_TOKEN} .` command which is safe for building docker images but it does not mention:~~
- ~~how to set local `NPM_TOKEN` variable~~
- ~~to clean terminal history after setting local `NPM_TOKEN` variable~~

~~Everyone seems to be setting their local `NPM_TOKEN` environment variable even going as far as putting it into `.zshenv` or `.bashrc` to have them always available during each terminal session which I think is an overkill and we should either promote a different approach or explain how to safely use the most common one.~~

~~I'm open to discussion and happy to turn this PR into a google searchable answer on how to safely build your docker images with personal access tokens as it seems the best-practices are not clear, or are not easily found or I haven't found them in which case - I'm happy to be shown the proper way of doing things or to be LMGTFYed.~~

<hr>

**edit**

I think there's a better approach and this section needs a big rewrite. The approach has been explained really well [here][Securely using NPM credentials with Docker].

Currently, npm docs encourages to use [`ARG` parameter  and `--build-arg` option](https://docs.docker.com/engine/reference/builder/#arg). When we follow the [link](https://docs.docker.com/engine/reference/builder/#arg) in the documentation we're going to end up reading a Warning and will be taken to [a better and newer solution][New Docker Build secret information] which does not use `ARG` but the global `.npmrc` file and which I've shown below as well.

I think this should be a more appropriate way of doing it:

1. Remember to login with `npm login` if you haven't already.

2. `Dockerfile`:
```Dockerfile
# https://docs.npmjs.com/docker-and-private-modules
FROM node:14

ENV APP_HOME="/app/"

WORKDIR ${APP_HOME}

COPY package*.json ${APP_HOME}/

RUN --mount=type=secret,id=npm,target=/root/.npmrc npm install

COPY . ${APP_HOME}/
CMD npm start
```

3. Build docker image with the command (on mac in this case)
```shell
docker build . -t secure-app-secrets:1.0 --secret id=npm,src=$HOME/.npmrc
```

If this solution finds some traction and acceptance I'm going to change this PR to include this solution but I feel like this would require a bigger re-write of this entire page if we're going to agree that this is the solution that should be promoted. 

My reasons for promoting it would be:
- Does not require setting of local `NPM_TOKEN` environment variable as the token is being taken from the global `.npmrc` which uses the token created during `npm login` phase and does not store it either
- No need to create local `.npmrc` and risk that someone will put their access token there and commit to source control
- No risk that someone will pass their access token to `docker build` command and make it visible in the docker image
- No need to worry about token getting into your terminal history
- [Seems to be using the latest docker features][New Docker Build secret information]

Again, happy to be shown a better way or maybe there is nothing wrong with the current solution. Either way, I'm opened to discussion and with the recent leak of [npm access token](https://thehackernews.com/2022/04/github-says-hackers-breach-dozens-of.html) I feel like it's an important one to have. 

## References
[New Docker Build secret information]: https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information
[Securely using NPM credentials with Docker]: https://medium.com/learnwithrahul/securely-using-npm-credentials-with-docker-f38b43bdb7c1

- [Using auth tokens in .npmrc](https://stackoverflow.com/questions/53099434/using-auth-tokens-in-npmrc)
- [How to set env var for .npmrc use](https://stackoverflow.com/questions/48728714/how-to-set-env-var-for-npmrc-use)
- [Docker build secrets and private npm packages](https://www.alexandraulsh.com/2019/02/24/docker-build-secrets-and-npmrc/)
- [Securely using .npmrc files in Docker images](https://www.alexandraulsh.com/2018/06/25/docker-npmrc-security/)
- [Build secrets and SSH forwarding in Docker 18.09](https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066)


content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -79,7 +79,11 @@ docker build --build-arg NPM_TOKEN=${NPM_TOKEN} .
 
 This will build the Docker image with the current `NPM_TOKEN` environment variable, so you can run `npm install` inside your container as the current logged-in user.
 
+To set your `NPM_TOKEN` environment varible run `export NPM_TOKEN="00000000-0000-0000-0000-000000000000"` and pass your npm access token instead of `0`s.
+
 <Note>
+  
+**Note:** After setting your `NPM_TOKEN` environment variable remember to clean your terminal history. You can do it with `history -c` command for `bash` and `history -p` for `zsh`.
```
ljharb, line 86: This is a dangerous command to be blindly recommending; terminal history is important.

matzar: Hey @ljharb could you please give some examples of when it's important? I'm not opposing. I'm genuinely interested.

matzar: It's me again @ljharb I've updated the PR as I think my current PR does not solve the issue but I've outlined a solution which I think might be what we need.

matzar:
I've made a proposition of changes to the documentation which would drop the usage of ARG and environment variables and would instead promote the usage of Docker build secrets.

matzar (commented):
The entire PR went in a different direction. Please have a look at the **edit** sections in the PR's description.

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
   - /private-modules/docker-and-private-modules
 ---
 
-To install private npm packages in a Docker container, you will need to use Docker's build-time variables.
+To install private npm packages in a Docker container, you will need to use [Docker build secrets](https://medium.com/learnwithrahul/securely-using-npm-credentials-with-docker-f38b43bdb7c1).
```
MylesBorins, line 7: Can we link to docker's official documentation here rather than a 3rd party blog?

I found https://docs.docker.com/engine/swarm/secrets/, but unsure if it is correct

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
...
 
-   ```
-   //registry.npmjs.org/:_authToken=${NPM_TOKEN}
-   ```
+The Dockerfile that takes advantage of this has a few more lines in it than the earlier example that allows us to use your global `.npmrc` and the access token created when running `npm login` command (if you haven't run it already - do so before moving on).
 
-   **Note:** that you are specifying a literal value of `${NPM_TOKEN}`.  The npm cli will replace this value with the contents of the `NPM_TOKEN` environment variable.  Do **not** put a token in this file.
+```dockerfile
+# https://docs.npmjs.com/docker-and-private-modules
+FROM node:14
```
MylesBorins, line 42:
```suggestion
FROM node:16
```

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
...
+This will configure your Dockerfile to receive `.npmrc` file via build secrets, that will leave no trace after npm dependency installation is done.
 
 ## Build the Docker image
 
 To build the image using the above Dockerfile and the npm authentication token, you can run the following command. Note the `.` at the end to give `docker build` the current directory as an argument.
 
-```
-docker build --build-arg NPM_TOKEN=${NPM_TOKEN} .
+```shell
+docker build . -t secure-app-secrets:1.0 --secret id=npm,src=$HOME/.npmrc
```
MylesBorins, line 65: should we be encouraging folks to be using their $HOME/.npmrc? I'd imagine they might be managing secrets specific for this deployment 🤔. In particular they might want to use a read-only token or automation token rather than the "full access" token that is used by default when logged in. This could also cause additional complications if 2FA is enabled on the account.

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
...
 ```
 
-This will build the Docker image with the current `NPM_TOKEN` environment variable, so you can run `npm install` inside your container as the current logged-in user.
+This will build the Docker image with the access token coming from your global `.npmrc` file received via build secrets, so you can run `npm install` inside your container as the current logged-in user.
```
MylesBorins, line 68: Again I think we should encourage folks to generate an access token for this use case rather than using the default token for an active CLI session

Jlemy29 (approved):

Jlemy29 (approved):

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
matzar: Committed the suggestion - thanks!

matzar: I actually think this would be a better reference link 
```suggestion
To install private npm packages in a Docker container, you will need to use [Docker build secrets](https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information).
```

matzar:
@ljharb @MylesBorins @trevrosen any chance on getting feedback on this please?

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
...
 
-## Update the Dockerfile
+WORKDIR ${APP_HOME}
 
-The Dockerfile that takes advantage of this has a few more lines in it than the earlier example that allows us to use the `.npmrc` file and the `ARG` parameter:
+COPY package*.json ${APP_HOME}/
 
-```
-FROM node
+RUN --mount=type=secret,id=npm,target=/root/.npmrc npm install
```
MylesBorins, line 50: in this example you are looking for `/root/.npmrc` correct? On most system the `.npmrc` is going to found in `~/.npmrc`.

Will this above command work for those installations out of the box?

matzar: > in this example you are looking for `/root/.npmrc` correct?

Yes, that's true.

I've only tested it on macOS and 
```ts
RUN --mount=type=secret,id=npmrc,target=~/.npmrc npm install
``` 
does not work.

AFAIA `/root/.npmrc` will resolve to `~/.npmrc` regardless of your operating system.

```ts
RUN --mount=type=secret,id=npmrc,target=.npmrc npm install
```
will also work and will take your `.npmrc` at the root if the local `.npmrc` is not present. 

Given that we want people to use the `.npmrc` at the root (`~`) where their npm token - generated with `npm login` - is, I think 
```ts
RUN --mount=type=secret,id=npmrc,target=/root/.npmrc npm install
```
is the way to go as it will force people to use the token generated with `npm login` and will not allow for silly behaviour when someone creates a local `.npmrc` and commits their npm token to the repo.

matzar (commented):
Changed the `id` on the RUN command to `npmrc`.

matzar (commented):
@MylesBorins I've also noticed a pretty bad error in the command itself, the `id` should be set to `npmrc`, not `npm`. Otherwise it will not work. I've commited a fix. Now it will work out-of-the-box.

content/integrations/integrating-npm-with-external-services/docker-and-private-modules.mdx
```diff
@@ -4,87 +4,71 @@ redirect_from:
...
+This will configure your Dockerfile to receive `.npmrc` file via build secrets, that will leave no trace after npm dependency installation is done.
 
 ## Build the Docker image
 
 To build the image using the above Dockerfile and the npm authentication token, you can run the following command. Note the `.` at the end to give `docker build` the current directory as an argument.
 
-```
-docker build --build-arg NPM_TOKEN=${NPM_TOKEN} .
+```shell
+docker build . -t secure-app-secrets:1.0 --secret id=npmrc,src=$HOME/.npmrc
```
matzar, line 65: @MylesBorins might have been a bad blunter but I've changed the `id` on the `docker build` command as well. We could reverse it back to `npm` or something else but I think `npmrc` might make a bit more sense?

matzar:
@MylesBorins @monishcm I think this is ready for another review. Appreciate you!

matzar (commented):
Looks good to me now.

matzar:
I've run `npm install` and pushed a commit with updated `package-lock.json` so all the test pass but it seems to have caused merge conflict which I can't solve w/out the write access (I'm not asking for ones - just saying :)). 

Would you rather not have `package-lock.json` in the PR @MylesBorins @ljharb @lukekarrys @spicycode ?

lukekarrys:
@matzar Yes, can you try removing the `package-lock.json` from this PR and see if the deployment succeeds?

matzar:
> @matzar Yes, can you try removing the `package-lock.json` from this PR and see if the deployment succeeds?

@lukekarrys I've removed `package-lock.json` and the deployment fails but running `gatsby build` succeeds. 

Maybe we should update `package-lock.json` on the main with another PR first?

MylesBorins:
@matzar can you try rebasing your PR against the `main` branch and force pushing. I'm not sure how you ended up in the situation you are in right now but `npm ci` is working locally without any conflicts

matzar:
> @matzar can you try rebasing your PR against the `main` branch and force pushing. I'm not sure how you ended up in the situation you are in right now but `npm ci` is working locally without any conflicts

@MylesBorins I've opened up a new PR #208 from the `main`, updated `docker-and-private-modules.mdx` with the changes that are in this PR, and the build passes (strangely enough, rebasing did not help).

I propose we close this PR and merge #208 instead which has all the same changes and is a lot less convoluted.

PR #107, opened 2022-04-27
Update about-packages-and-modules.mdx

yoyoyojoe:
Straightforward edit to improve formatting consistency for a section within [About packages and modules](https://docs.npmjs.com/about-packages-and-modules).

## Currently

This looks **_off_**. 

image

## Proposed Edit

I believe this is the originally intended format. It looks consistent with the remainder of the page as well.

image

## References

See above notes.


wraithgar (approved):
This definitely looks like an automated cleanup gone wrong. Good catch, thank you.

PR #111, opened 2022-05-02
chore: tweak update-cli workflow a bit to remove the deploy key

nlf:
there were only minor differences between the publish and update-cli workflows so the most direct route to removing the need for the deploy key here was to add the second push to the `dist` branch

update:

i also added a commit that rotates the secret in use for the docs-staging deploy and another updating the bot account information in all of the workflows

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: [ main ]
 
+permissions:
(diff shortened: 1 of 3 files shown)


MylesBorins:
adding @ethomson for a review, who originally wrote this workflow

wraithgar (approved):

PR #127, opened 2022-05-25
doc: password reset no longer revokes tokens

MylesBorins:
This was the only place it was documented. We've updated this behavior
as we no longer have the same risk to tokens on password reset with
login verification


nishantms (approved):

nishantms:
LGTM 🚀

PR #160, opened 2022-07-29
tools: do pages deployment via actions

MylesBorins:
There is now support for deploying pages by actions bypassing the need to
upload to a branch, this significantly simplifies the overall process
and removed the need for a "dist" branch.

I'm about 99% confident this will work out of the box, I've tested on
two personal websites already.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,27 +1,29 @@
 name: Publish
 
+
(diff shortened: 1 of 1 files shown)


ronniephlee76 (commented):
#

lukekarrys (approved):

PR #184, opened 2022-08-25
Added a brief expansion of the term npm

julianharty:
I wanted to reference npm where my readers would understand what npm stands for (represents). I found this about page and discovered it doesn't say what npm stands for (unlike https://www.w3schools.com/whatis/whatis_npm.asp#:~:text=The%20name%20npm%20(Node%20Package,json.)


wraithgar:
That is not what npm stands for: https://github.com/npm/cli#is-npm-an-acronym-for-node-package-manager

PR #191, opened 2022-08-29
fix dependencies Command

kalanakt:
- [x] fix issue in dependencies command


MylesBorins:
Can you expand on what the issue is? The current text makes it clear the flag is optional

kalanakt:
feels like ` npm install <package-name> [--save-prod] ` not friendly command line

ljharb:
The `< >` means required, and the `[ ]` means optional, and neither are supposed to be typed directly into the command line here.

PR #216, opened 2022-09-26
docs: Add note to docker-and-private-modules about security

destumme:
* Adding note about security implications of using docker --build-arg


destumme:
I wasn't sure what y'alls preference for cross linking docs was. So here is the link for more info about buildkit: https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information I can add a link to the note if you'd like. 

I hope this was a welcome contribution, apologize if I should have opened an issue or followed another process, seemed like small enough change to just go for it.

wraithgar:
There is [another pr](https://github.com/npm/documentation/pull/208) in the works that does not suggest that flag at all.

PR #239, opened 2022-10-08
feat: build cli content from registry

lukekarrys:
This change will allow the CLI content to be built from the docs
published in the registry tarball. This is necessary now that the CLI
repo will no longer contain the generated docs in source countro. There
still exists a fallback for fetching from GitHub for legacy versions of
the CLI that have more recent content in a branc that was never
published to the registry.

Because of this change, the CLI is no longer checked in as submodules.
In order to avoid fetching more often than necessary, a `resolved` field
is stored in the `releases.json` manifest.

This commit also introduced `@npmcli/template-oss` as a dev dep to the
CLI workspace to standardize the development workflow.


timscharfenort8989 (commented):
Gut

PR #253, opened 2022-10-14
Add a title to the homepage link

anurbol:
Added a title to the homepage link. 

## References
  Related to https://github.com/github/accessibility-audits/issues/2739

### DIFF
--- a/theme/src/components/header.js
+++ b/theme/src/components/header.js
@@ -45,7 +45,7 @@ function Header({location, isSearchEnabled = true}) {
         bg="#333333"
       >
         <Flex alignItems="center">
-          <Link as={GatsbyLink} to="/" style={logoStyle} mr={3}>
+          <Link as={GatsbyLink} to="/" style={logoStyle} mr={3} title='NPM home'>
             <NpmLogo size="32" />
           </Link>
           <Link as={GatsbyLink} to="/" style={titleStyle} mr={4}>
--- a/theme/src/components/header.js
+++ b/theme/src/components/header.js
@@ -31,8 +31,8 @@ function Header({location, isSearchEnabled = true}) {
   const [isMobileSearchOpen, setIsMobileSearchOpen] = React.useState(false)
   const siteMetadata = useSiteMetadata()
 
-  const logoStyle = { 'color': '#cb0000' };
-  const titleStyle = { 'color': '#dddddd', 'fontWeight': '600' };
+  const logoStyle = { 'color': '#cb0000', 'marginRight': '16px' };
+  const titleStyle = { 'color': '#dddddd', 'fontWeight': '600', 'display': 'flex', 'alignItems': 'center' };
 
   return (
     <Sticky>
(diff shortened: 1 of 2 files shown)


anurbol:
I assessed a couple of other options as well:

- Adding `<title>` to the svg doesn't feel correct as the [issue](https://github.com/github/accessibility-audits/issues/2739) says we need to add title to the link and not to the image. 
- Adding `alt` attribute is inappropriate as it is intended for using on `img`

20shivangi (commented):
Can you please attach a vide of voiceover on Safari with this change once ?

prabum3:
We already have a link to `npmjs.com` on the right side. I think it is fine to have the logo redirect to docs homepage and have the title `npm docs`.

theme/src/components/header.js
```diff
@@ -45,7 +45,7 @@ function Header({location, isSearchEnabled = true}) {
         bg="#333333"
       >
         <Flex alignItems="center">
-          <Link as={GatsbyLink} to="/" style={logoStyle} mr={3}>
+          <Link as={GatsbyLink} to="/" style={logoStyle} mr={3} title='NPM home'>
```
20shivangi, line 48: Please change the title to `npm Docs`

Also, npm should always be lowercase and not NPM :  https://github.com/npm/cli#faq-on-branding

amit-avit:
+1 to the comments to just make it as `npm Docs`, also do check once that the tool passes with updating the `title` else we might have to update `aria-label` instead.

amit-avit:
@anurbol @prabum3 @20shivangi  It seems we have 2 `a` pointing to the docs homepage one for logo another for text `npm Docs`. I would suggest that we only have one `a` and move the logo inside the `a` and then the text would act as title and the logo would be aria-hidden.

20shivangi:
> @anurbol @prabum3 @20shivangi It seems we have 2 `a` pointing to the docs homepage one for logo another for text `npm Docs`. I would suggest that we only have one `a` and move the logo inside the `a` and then the text would act as title and the logo would be aria-hidden.

Yes @amit-avit , that's what I was saying that there will be duplication :D
We can go with this approach

anurbol:
@amit-avit @prabum3 @20shivangi 

I've merge the two links into one (there is now single `a` instead of two). Because previously there were two independent elements - logo and title, they were positioned on an arbitrary height relative to each other, and now they are strictly vertically aligned in their common link element (which lead to the logo now being positioned a bit lower). Which I believe is a good thing. But if I am mistaken please let me know.

https://user-images.githubusercontent.com/3603448/196427424-9af4d6a0-769c-4394-a944-996cf2bbf7df.mp4

amit-avit (approved):
LGTM
@anurbol personally I feel the alignment is better now with your changes.

```
     _ 
    /|\      |   |
   /_|_\     | __|__
 ____|____   |   |
 \_o_o_o_/   |   \__/

```

theme/src/components/header.js
anurbol: Please see https://github.com/npm/documentation/pull/253#issuecomment-1282300049

20shivangi (approved):

Seryozha95 (approved):

Seryozha95 (approved):

PR #262, opened 2022-10-19
Announce copy to clipboard action

anurbol:
Fix screen reader not announcing any information upon activating 'Copy to Clipboard' button

## References
Related to https://github.com/github/accessibility-audits/issues/2795

### DIFF
--- a/theme/src/components/clipboard-copy.js
+++ b/theme/src/components/clipboard-copy.js
@@ -2,6 +2,7 @@ import {Button, StyledOcticon} from '@primer/components'
 import {CheckIcon, CopyIcon} from '@primer/octicons-react'
 import copy from 'copy-to-clipboard'
 import React from 'react'
+import Notification from './notification'
 
 function ClipboardCopy({value}) {
   const [copied, setCopied] = React.useState(false)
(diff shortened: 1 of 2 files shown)


anurbol:
Considered options:

- [Focus a hidden element with the announcement text](https://github.com/npm/documentation/pull/262/commits/433e5993b93983fa9e61c9a7608a93337920f461). Doesn't work nice for Safari tried workarounds, not working and it is becoming hackish and ugly.
- Create a notification/toast element and style it GitHub way (https://primer.style/css/components/toasts) - didn't figure out how to use the `.Toast` class, didn't find any usage across wubwub and documentation projects.
- [Primer React](https://primer.style/react/) doesn't seem to have a notification/toast component. The closest thing to that is the Dialog. Customizing it would require a lot of effort, IMO.
- Gatsby seems to have a notification feature, but it is probably only for system/its own usage - couldn't find a way to use it from a theme. Gatsby plugins seem to be about build time notifications and not web notifications.

anurbol:
I now want to make a notification component like in wubwub. Since it seems it is the only proper way to solve issue.

PR #263, opened 2022-10-19
Aria label associated

makhmudjon-inadullaev:
Users who depend on screen readers would be impacted and might fail to access the information about the version dropdown label, if a separate label is not provided for the combo box describing its purpose.

image

## References

Issue: https://github.com/github/accessibility-audits/issues/2748


Seryozha95 (approved):

makhmudjon-inadullaev (commented):
aria label moved to <summary />

PR #270, opened 2022-10-22
Default {prefix} on Windows should be %ProgramFiles%\nodejs.

jhribal:
Hello,

i changed the default {prefix} value on Windows because current doc says it is **`%AppData%\npm`** but globally installed packages are by default stored under `%ProgramFiles%/nodejs/node_modules` so {prefix} should be **`%ProgramFiles%/nodejs`**.

**What current doc says:**
> prefix Configuration
> The prefix config defaults to the location where node is installed. On most systems, this is /usr/local. **On Windows, it's %AppData%\npm.** On Unix systems, it's one level up, since node is typically installed at {prefix}/bin/node rather than {prefix}/node.exe.

### DIFF
--- a/content/cli/v6/configuring-npm/folders.md
+++ b/content/cli/v6/configuring-npm/folders.md
@@ -30,7 +30,7 @@ This document will tell you what it puts where.
 #### prefix Configuration
 
 The `prefix` config defaults to the location where node is installed.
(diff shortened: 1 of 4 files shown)


jhribal:
I closed this PR because when node.js is installed using installer from web the prefix is really %AppData%\npm. I had my node.js installed using nvm for windows.

PR #273, opened 2022-10-24
Changed placeholder text

pavlov-kirill:
Screen Shot 2022-10-24 at 15 11 24

## References
https://github.com/github/accessibility-audits/issues/2744


20shivangi (approved):

amit-avit (approved):
I personally feel mobile view can have shorter text like this and it seems to be intentional but I don't have any strong feeling for either so I'll let it go.

PR #296, opened 2022-10-31
Make scrollable code boxes accessible by keyboards

anurbol:
todo

## References
Related to https://github.com/github/accessibility-audits/issues/2746

### DIFF
--- a/theme/src/components/code.js
+++ b/theme/src/components/code.js
@@ -41,6 +41,7 @@ function Code({className, children, live, noinline}) {
                     key={key}
                     fontFamily="mono"
                     fontSize={1}
+                    tabIndex={0}
                     {...getTokenProps({token, key})}
                   />
                 ))}


anurbol:
Screen Recording (Ubuntu, Chrome)
https://user-images.githubusercontent.com/3603448/199412625-54e9cd24-2bfa-4a41-9461-ac0355e48342.mp4

Also tested in Firefox (Ubuntu), Safari (MacOS)

theme/src/components/code.js
```diff
@@ -41,6 +41,7 @@ function Code({className, children, live, noinline}) {
                     key={key}
                     fontFamily="mono"
                     fontSize={1}
+                    tabIndex={0}
```
amit-avit, line 44: This change is making the screen reader weird.

amit-avit: But adding tabIndex to the pre tag works fine but since it is inside external component we won't be able to make such change. 😞 

image

anurbol: @amit-avit exactly, it's a 3rd part component... Can you please elaborate on what is exactly weird with the screen reader?

amit-avit: If u see in the attached video for the focusable box, voiceOver reads the content 3 times.

https://user-images.githubusercontent.com/8944724/199778850-832936f4-f803-4acb-ba50-b8eff2152935.mov

anurbol: Unfortunately I cannot reproduce it neither in Safari nor in Chrome, on MacOS. I have a theory, at it looks like it judjing by your video that you have some additional Chrome extension that causes such behavior. Can it be the case? On clean chrome it reads exactly one time. 

P.S. What app are you using to screen record on your MacOS? cmd+shift+5 doesn't work properly with VoiceOver enabled for me...

amit-avit: I'm able to repro in chrome but not in safari. I hope your are trying to navigate using ctrl+opt+righArrow and not tab.

anurbol: Yep, that combo helped to reproduce. BTW is it the correct way to navigate for a11y testing?

anurbol: Ok, thanks, I will use it to test the fix to this triple announcement.

anurbol: Hey @amit-avit I just got back on this today, and seem to have fixed it, can you please confirm?

anurbol:
Finally fixed the issue on MacOS+Chrome

https://user-images.githubusercontent.com/3603448/201635533-ae02bd75-669a-4901-8018-e5621f819f1c.mp4

amit-avit:
I'm ok with the idea behind the fix but when the new div gets focused the outline looks weird.
image


I would say we fix this as well before we merge this.

anurbol:
> 



> 



> I'm ok with the idea behind the fix but when the new div gets focused the outline looks weird. image
> 
> I would say we fix this as well before we merge this.

I made a few experiments, but none of them seems to be nice enough. 

1) We could hide the outline - but then the user would be confused where did the focus go.
2) We could apply `display: block; width: 100%;`, but then the outline will be a black wide line, which is also weird, isn't it? 

We cannot put the code contents into the div with `tabindex`, this will cause MacOS' Chrome read the code three times (as it will consider it a "group" - not sure why)

Seryozha95:
> > 
> 
> > 
> 
> > I'm ok with the idea behind the fix but when the new div gets focused the outline looks weird. image
> > I would say we fix this as well before we merge this.
> 
> I made a few experiments, but none of them seems to be nice enough.
> 
> 1. We could hide the outline - but then the user would be confused where did the focus go.
> 2. We could apply `display: block; width: 100%;`, but then the outline will be a black wide line, which is also weird, isn't it?
> 
> We cannot put the code contents into the div with `tabindex`, this will cause MacOS' Chrome read the code three times (as it will consider it a "group" - not sure why)

@anurbol `visibility: hidden` should fix the focus issue.

anurbol:
> 

Unfortunately no, it prevents it from getting focus at all.

amit-avit:
Can we try to make this new div have the width and height as the parent container and make use of z-index to move it behind the text elements

anurbol:
> 



> Can we try to make this new div have the width and height as the parent container and make use of z-index to move it behind the text elements

I think this can't be done with CSS alone, but with the help of javascript. Okay I'll try

theme/src/components/code.js
amit-avit: Yes this is fixed.

amit-avit:
@arisacoba a fix is made to make a text box with horizontal scrollbar to be able to scroll while using keyboard. When the element is focused it looks as below, is it ok to proceed with?
image

anurbol:
@amit Done [here](https://github.com/npm/documentation/pull/296/commits/144f164539ce9c3593489b1445dedfe73c68d011). Tried to make it as simple as possible... still doesn't look very nice

amit-avit (approved):
This looks much better even though a bit more complex.
LGTM

BTW you tagged wrong `amit` in ur comment 😁

arisacoba:
@anurbol @amit-avit best to cross-post this concern with `#accessibility-design` on Slack if we haven't yet. They have the best knowledge of how this can be implemented properly

Seryozha95 (approved):

lindseywild:
Hi! Trying to review this using the instructions in the README but `npm run develop` errors out -- do you know if there's an updated way to view this locally, or a staging URL I can review? Thanks!

20shivangi:
> Hi! Trying to review this using the instructions in the README but `npm run develop` errors out -- do you know if there's an updated way to view this locally, or a staging URL I can review? Thanks!

@lindseywild You can make use of this preview URL : https://npm-f9a17b3f5d-15913497.drafts.github.io/, it has the changes that are made in this PR

lindseywild:
Thank you for this! It's a lot better than what is currently there, which doesn't work at all, so I like it! I realize that this isn't entirely elegant due to the constraints of the third-party code. I was able to test this on Chrome with a keyboard, Safari with a keyboard + VoiceOver, and NVDA + Chrome. NVDA reads "blank" when tabbing to the code example, but I'm not sure if that can be easily fixed... It doesn't prevent the user from doing anything, it's just a little odd. To cover our bases, if you could open an issue in the third-party library stating that keyboard functionality doesn't work for scrollable boxes, that'd be great -- that will allow us to close the audit issue even with a few quirks. Thanks!

Spiritaine:
Awesome thank you

PR #328, opened 2022-11-30
What should be the recommended way to deal with multiple versions of Node?

cinderblock:
I recently learned that Npm documentation is **recommending** the use of `npx` to manage versions of Node.js. I, humbly, think this is a **bad** recommendation as it is an anti-pattern and is specifically is confusing to users, new and old. It also, simply, doesn't directly work on Windows (and won't ever).

The permission issues, as I understand them, are standard issues with installing into system locations. These could be solved the same way many other popular programs have solved it, by having a "System Installer" and a "User Installer". In either case, installing packages "globally" is generally discouraged and only specific situations are helped by having a global version installed.

If managing versions is important to certain users, I think there is also a more portable and reliable way to have that version pinned. Ignoring `package.json#engines`, you can simply install `node` as an `npm` dependency and, with few exceptions, everything "Just works" as you want it to.

### DIFF
--- a/content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
+++ b/content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
@@ -3,7 +3,7 @@ title: Downloading and installing Node.js and npm
 redirect_from: [ /getting-started/installing-node ]
 ---
 
-To publish and install packages to and from the public npm registry or a private npm registry, you must install Node.js and the npm command line interface using either a Node version manager or a Node installer. **We strongly recommend using a Node version manager like [nvm](https://github.com/nvm-sh/nvm) to install Node.js and npm.** We do not recommend using a Node installer, since the Node installation process installs npm in a directory with local permissions and can cause permissions errors when you run npm packages globally.
+To publish and install packages to and from the public npm registry or a private npm registry, you must install Node.js and the npm command line interface using either a Node version manager or a Node installer.
 
 <Note>
 
@@ -18,6 +18,7 @@ npm install -g npm
 ## Overview
 
 - [Checking your version of npm and Node.js](#checking-your-version-of-npm-and-nodejs)
+- [Using Npm to install Node locally](#using-npm-to-install-node-locally)
 - [Using a Node version manager to install Node.js and npm](#using-a-node-version-manager-to-install-nodejs-and-npm)
 - [Using a Node installer to install Node.js and npm](#using-a-node-installer-to-install-nodejs-and-npm)
 
@@ -29,6 +30,10 @@ To see if you already have Node.js and npm installed and check the installed ver
 node -v
 npm -v
 ```
+
+## Using Npm to install Node locally
+
+You can `npm i node@version` to install a specific version of node. Npm will use the installed version when running any scripts defined in `package.json#scripts`. This is recommended because it lets you lock a particular project to a specific version of Node without affecting the rest of your system
 
 ## Using a Node version manager to install Node.js and npm
 


content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
```diff
@@ -30,6 +31,10 @@ node -v
 npm -v
 ```
 
+## Using Npm to install Node locally
+
+You can `npm i node@version` to install a specific version of node. Npm will use the installed version when running any scripts defined in `package.json#scripts`. This is recommended because it lets you lock a particular project to a specific version of Node without affecting the rest of your system
```
MylesBorins, line 36: This is not a way to install the Node.js that I would like us advocated as npm.

MylesBorins:
Can you share a link to permissions issues being solved with the node.js installer? The warning likely needs to remain until this permission issue is fixed on all LTS version's of node.js

content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
cinderblock: I guess I don't understand how this way is at all worse than the suggested `nvm` way. It's easier, faster, more reliable, and reproducible. `nvm` doesn't run on Windows. What does `nvm` have that `npm i --save node` doesn't?

Or, at least, why is "nvm" the "strongly recommended" way? I see the explanation about npm permissions but that's only a problem if you're installing global packages, which isn't the default and requires elevated privileges anyway. If it's not recommended, why is this the default Node.js has stuck with? I missing something?

cinderblock:
> Can you share a link to permissions issues being solved with the node.js installer? The warning likely needs to remain until this permission issue is fixed on all LTS version's of node.js

I guess I'm not sure what permissions this is even referencing. I've **never** run into them and **always** installed with the official Node.js installer, on Windows and Linux. Can you give an example of the permission issue that this recommendation is trying to avoid?

content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
sophacola: > I guess I don't understand how this way is at all worse than the suggested `nvm` way. It's easier, faster, more reliable, and reproducible. `nvm` doesn't run on ~~Windows~~ non-POSIX. What does `nvm` have that `npm i --save node` doesn't?
> 
> Or, at least, why is `nvm` the "strongly recommended" way? I see the explanation about Npm permissions but, as far as I understand, that's only a problem if you're installing global packages, which isn't the default and requires elevated privileges anyway. If it's not recommended, why is this the default Node.js has stuck with? I missing something?

I worked it out I think

MylesBorins: installing Node.js has some shortcomings such as "npm install" and initial npm scripts running with a different version of npm / node.js then subsequent scripts without making changes to your system path.

If you try to run `node` or `npm` your system will first attempt to use the version in your path. It is only npm scripts that inject `node_modules/.bin` into your path. So if you were to do something like run `npm test` that will initially use your system version of node.js / npm.

These kinds of inconsistencies open up for a large number of possible edge cases and inconsistencies

MylesBorins:
The Node.js installer on MacOS installs into /usr/local which by default on MacOS is owned by root. This means that unless you chown the directory, something that is dangerous to do if you don't know what you are doing, all global installs will fail without using `sudo`. Running with sudo is dangerous from a permissions perspective, meaning all postinstall scripts in all modules are running as root. Also if you do ever run `sudo npm install` then your cache begins to become populated with packages that are owned by root which result in future installs failing, which for many new developers means "running sudo npm install" all the time.

There are some similar issues with windows permissions with the installer, but I'm less familiar with those systems. In general for this type of instruction it is better to have a single blessed path instead of multiple options. 

I am on the Node.js Release team and there have been many discussions about trying to find a way to improve this default installation but it is a non-trivial task the project hasn't found resources to support.

cinderblock:
> The Node.js installer on MacOS installs into /usr/local which by default on MacOS is owned by root. This means that unless you chown the directory, something that is dangerous to do if you don't know what you are doing, all global installs will fail without using `sudo`.

On Linux it is installed to `/usr/bin`. I would never suggest making such directories writable to all.

> Running with sudo is dangerous from a permissions perspective, meaning all postinstall scripts in all modules are running as root.

This is **only** applicable when you `sudo npm i -g <package>`, which is never recommended. (or shouldn't ever be, imho)

> Also if you do ever run `sudo npm install` then your cache begins to become populated with packages that are owned by root which result in future installs failing, which for many new developers means "running sudo npm install" all the time.

So, because some developers use `sudo` when they shouldn't, all developers are suggested to use a third party package that just does everything in a user's directory?

How does the `nvm` method prevent users from doing exactly this anyway?

> There are some similar issues with windows permissions with the installer, but I'm less familiar with those systems.

These only seem to be true when users try to install global modules, something that isn't recommended.

> In general for this type of instruction it is better to have a single blessed path instead of multiple options.

I don't understand why the "single blessed path" is the one that it currently is. `nvm` does not work on Windows, so it's already two paths. It seems to me telling people to not mess up permissions would be a more consistent single path. Npm could even verify that permissions are what they're expected to be if this is that much of an issue.

Ideally, shouldn't `node` "just work"? Especially with Node-API maintaining binary compatibility between packages, managing the current version of `node` **should** be a thing of the past. Sure, there are plenty of packages that haven't switched, but those seem to be per-package issues.

Sure, there are breaking changes in `node` that happen, but if you've only ever had one version of node installed, that won't be a problem either.

> I am on the Node.js Release team and there have been many discussions about trying to find a way to improve this default installation but it is a non-trivial task the project hasn't found resources to support.

I really don't see what the issue is. It's pretty normal to require an admin to install programs on a computer and not let users mess with system files. I don't see these issues as issues. It seems like things are working as designed.

I'm wondering how other programs deal with this issue. VS Code, for instance, provides a "User Installer" version (by default) but also provides a "System Installer" version.

content/getting-started/configuring-your-local-environment/downloading-and-installing-node-js-and-npm.mdx
cinderblock: @MylesBorins That is a good point, but is a known issue. This only affects the *first* compilation on every system which is rather manageable, imho. It also only affects repos that "opt-in" to using the `node` package on Npm. Usually we don't need to lock `node` to a particular version, or at least `package.json#engines` is usually adequate in my experience. Note, once installed, `npm test` **will** use the correct version of `node` for actually **running** the tests. The top level node that is running npm may be different, but that is *desired*. What if I want to use modern versions of Npm that only work on recent versions of Node but run a project with an older version of Node?

This also ignores the issues with `nvm`

- It's a 3rd party package (yes, the `node` package is as well, but not everyone needs it)
- It doesn't work on Windows directly (without WSL or msys or needs `nvm-windows`...)
- It affects the other projects (I can't easily work on two projects with different node versions at the same time and messing with PATH is something I think should be avoided, especially on Windows)
- Needs extra configuration (`.nvmrc`) to ensure correct version is easily installed

MylesBorins: > that is a good point, but is a known issue. This only affects the first compilation on every system which is rather manageable, imho

This is where we disagree. This is not an acceptable trade off in my opinion and I don't think we should be suggesting this solution in the npm documentation.

MylesBorins (changes requested):
I'm not in support to these changes to our documentation. I think it is very reasonable to advise individuals to use a node version manager rather than the node.js official installer. I also do not think that we should be advocating for the use of the node package (which has a variety of edge cases).

I think that we could consider documenting additional version management solutions, especially documenting something that can be supported on windows.

For what it is worth I would similarly not advise using an operating system's built in package manager such as apt.

This opinion is derived from years of working with developers of a variety of skill levels, including mentoring a number of new developers at node-school.

I don't come to the above conclusion lightly. The installation of Node.js is in general requires a significant overhaul but the project hasn't been able to invest time in it. As a member of the Node.js project I have contributed to both the Windows + MacOS installer, have worked to help make a platform agnostic electron based installer, and helped lead sessions at the collaborator summit to help think about "next generation" ways to think about managing both Node.js + package managers on systems.

cinderblock:
> This is where we disagree. This is not an acceptable trade off in my opinion and I don't think we should be suggesting this solution in the npm documentation.

I don't really think either should be recommended in the Npm documentation. We should recommend using the current stable version of Node.

Personally, I think scripts messing with what happens when I type `node` in my terminal is not acceptable. I expect `node` to run the current system version of Node.js.

> I'm not in support to these changes to our documentation. I think it is very reasonable to advise individuals to use a node version manager rather than the node.js official installer. 

I don't understand not recommending the use of the latest stable version with the official installer. I would expect "stable" and "official" are always what should be recommended.

> I also do not think that we should be advocating for the use of the node package (which has a variety of edge cases).

I have the same feeling about `nvm`, except that `nvm`'s headaches and pitfalls are, imho, worse.

> I think that we could consider documenting additional version management solutions, especially documenting something that can be supported on windows.

A reasonable compromise might be listing the `nvm` solution alongside the `npm i node` solution, with their caveats.

> For what it is worth I would similarly not advise using an operating system's built in package manager such as apt.

I don't agree with this at all. Maybe this is our fundamental disagreement.

> This opinion is derived from years of working with developers of a variety of skill levels, including mentoring a number of new developers at node-school.

So, because some students seem to be comfortable with this method, all Npm users should be recommended to use `nvm`, and only that? I'm sorry to hyperbolize your comments but I don't think this is a good enough reason. I have also mentored some new learners and we have never had version issues. I would be curious to see an objective measurement comparing these methods, but that may be a little beyond what is possible.

Looking at NodeSchool, I see that all of the courses recommend installing with `-g`, so maybe this is where your permission experiences come from? I suspect things would be easier if the `-g` were dropped and `npx` used instead. I see that `npx` wasn't standard when NodeSchool started recommending this option. In any case, I think that's not a discussion for here.

> I don't come to the above conclusion lightly. The installation of Node.js is in general requires a significant overhaul but the project hasn't been able to invest time in it. As a member of the Node.js project I have contributed to both the Windows + MacOS installer, have worked to help make a platform agnostic electron based installer, and helped lead sessions at the collaborator summit to help think about "next generation" ways to think about managing both Node.js + package managers on systems.

I love electron but this actually worries me. Each OS has time honored standard ways to install programs - breaking them would, imho, cause more headaches than they solve, for developers and users.

What do you think about having two versions of the installer like VS Code does? Or a checkbox when installing to select a system location vs user location?

MylesBorins:
> What do you think about having two versions of the installer like VS Code does? Or a checkbox when installing to select a system location vs user location?

I think that is out of scope for the npm documentation repo.

> So, because some students seem to be comfortable with this method, all Npm users should be recommended to use nvm, and only that? 

I think I did a poor job of communicating my point, which was not about students or node school. My current position comes from years of industry experience working with a variety of developers ranging from student to enterprise. As a member of the Node.js project I've been dealing with release management and version management of Node.js for close to a decade. I'm not trying to flaunt my laurels, more so I want to be clear that my position is based on extensive experience and not very likely to change in this PR based on debate.

The official Node.js installers are not a way that we will be advocating to install Node.js based on extensive experience with system issues and known pitfalls with their design. Installing Node.js via a version manager, such as nvm (which is an OpenJS project) is in the opinion of myself, and the npm project, the preferred way to manage Node.js on a system.

A path forward I can see for this PR is to expand the list of version managers as opposed to removing npm. I will not be convinced that the node package should be recommended or ever used, I would object to them being included in the npm docs.

MylesBorins:
> > For what it is worth I would similarly not advise using an operating system's built in package manager such as apt.

> I don't agree with this at all. Maybe this is our fundamental disagreement.

Node.js follows an LTS program that doesn't align with the philosphy of most linux distros such as apt. It also vendors all it's own dependencies statically, which goes against the dynamic dependency philosphy of a OS like APT. You are pretty much guaranteed to get a version of Node.js that is old and built differently than what is extensively tested by the Node.js project if you use an operating systems package manager... that is why a version manager is suggested for local dev and the official docker images for production.

I apologize but I really have limited time and cannot continue to debate all the finer points that are being brought up in this PR. I appreciate you taking the time to contribute but I hope you can understand and respect that I'll need to limit my ongoing participation here to review of the contents and won't be engaging much more in the broader discussion.

lukekarrys:
I agree with @MylesBorins feedback and will be closing this PR since it won't be landed in it's current form.

If anyone would like to continue work on this, here is the relevant feedback from @MylesBorins about a possible path forward:
> A path forward I can see for this PR is to expand the list of version managers as opposed to removing npm. I will not be convinced that the node package should be recommended or ever used, I would object to them being included in the npm docs.

PR #329, opened 2022-11-30
Fixed Typo update  Command | npm-update

nishantsikarwar:
The aliases mentioned in the rpm-update page is incorrect 
https://docs.npmjs.com/cli/v9/commands/npm-update

```
npm update [<pkg>...]

aliases: up, upgrade, udpate
```

### DIFF
--- a/content/cli/v9/commands/npm-update.md
+++ b/content/cli/v9/commands/npm-update.md
@@ -40,7 +40,7 @@ redirect_from:
 ```bash
 npm update [<pkg>...]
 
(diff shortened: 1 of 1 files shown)


lukekarrys:
`udpate` is not a typo, it is allows the `update` command to be run even if it is mistyped as `udpate`.

The other one is a typo, but will need to be fixed in the `npm/cli` repo here: [udpate](https://github.com/npm/cli/edit/latest/docs/lib/content/commands/npm-update.md)

PR #331, opened 2022-12-01
Changed `OS X` to `macOS`

arunsingh-me:
Documentation uses older name of Mac operating system i.e. no longer being used.


hexiaokang (commented):
o(￣▽￣)ｄ

PR #371, opened 2022-12-25
Fix italic formatting in link text

ericmutta:
Markdown doesn't seem to support direct formatting of raw links, so the current text produces this broken output where the `*` symbols leak through and the italic formatting is applied incorrectly:

image

The fix turns the raw link into a markdown link then applies the italic formatting to the link text, to produce this:

image


lukekarrys:
This is how it is displaying for me in the deployed preview (https://npm-d2d6eb9157-15913497.drafts.github.io/creating-and-publishing-unscoped-public-packages):

Screen Shot 2023-01-01 at 7 09 33 PM


Are you seeing the same thing at the preview url above?

ericmutta:
Hi @lukekarrys happy new year! And many thanks for following up on this PR!

I can confirm that I see the same thing you see on the preview URL you gave above (i.e. the the backslash is leaking through and being displayed when it shouldn't).

Please note that I am doing all these edits exclusively on GitHub (i.e. I did not clone the repo locally) and when I do a preview of the edit on the GitHub editor, here is what I see:

image

I think you may be able to preview the edits in the same way via this link: https://github.com/npm/documentation/blob/d184c8b0daa67f7dbad7cd8b7557aaa40831fc60/content/packages-and-modules/contributing-packages-to-the-registry/creating-and-publishing-unscoped-public-packages.mdx

If it proves too fidgety I suppose we can just leave the link as is, then italicize the "...replacing _package-name_ with..." portion (let me know if I should do that) :+1:

ericmutta:
Another example of broken italic formatting in link text from here https://docs.npmjs.com/about-npm#getting-started

image

The original markdown looks like this:

```
...which will be available at http://www.npmjs.com/~*yourusername*.
```

Whatever fix we ultimately implement in this PR will need to be applied to that page too :+1:

Khalidalmasri:
تم الإرسال من Galaxy الخاص بي
-------- الرسالة الأصلية --------من: Eric Mutta ***@***.***> التاريخ: ٥‏/١‏/٢٠٢٣  ٢:٣٣ ص  (GMT+03:00) إلى: npm/documentation ***@***.***> نسخة: Subscribed ***@***.***> الموضوع: Re: [npm/documentation] Fix italic formatting in link text (PR #371) 

Another example of broken italic formatting in link text from here https://docs.npmjs.com/about-npm#getting-started



The original markdown looks like this:

...which will be available at http://www.npmjs.com/~*yourusername*.



Whatever fix we ultimately implement in this PR will need to be applied to that page too 👍

ericmutta:
Thanks for finally getting round to this @lukekarrys, looks like using backticks did the trick 👍

PR #387, opened 2023-01-02
fix: change publish workflow to pull_request

lukekarrys:


.github/workflows/publish.yml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
```
ljharb, line 9: it won't have access to secrets if you make this change.

lukekarrys: but only from 3rd party forks right? the build is designed to gracefully degrade without the secret so it just wont pull in some github specific data for the preview deployment

PR #428, opened 2023-02-01
change in grammar

vinysinza:
We do not recommend using a Node installer, since the Node installation process installs npm in a directory with local permissions and can cause permission errors when you run npm packages globally.


wraithgar:
[permissions errors](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) is the correct term and is consistently used that way across our docs.

PR #430, opened 2023-02-01
doc: remove reference to lastpass

MylesBorins:
based on the [recent security incident](https://blog.lastpass.com/2022/12/notice-of-recent-security-incident/) I don't think we should still be recommending lastpass


Mohammedfaha (commented):
Ksa.saudiarabic.sa@gmail.com

PR #431, opened 2023-02-02
Converting all component `.js` files to `.jsx` files for accessibility linting

20shivangi:


kendallgassner (commented):
We also might want to install this plugin:

- https://github.com/github/eslint-plugin-github

PR #436, opened 2023-02-07
Update about-packages-and-modules.mdx

D7x7z49:
Abnormal rendering of changed part in browser


lukekarrys:
This was fixed in #460

PR #469, opened 2023-03-13
Update default branch name from master to main

seantcoyote:
This proposed PR updates a reference to the default branch name from `master` to `main`.

Reference: https://github.com/github/renaming

### DIFF
--- a/content/packages-and-modules/introduction-to-packages-and-modules/about-packages-and-modules.mdx
+++ b/content/packages-and-modules/introduction-to-packages-and-modules/about-packages-and-modules.mdx
@@ -37,7 +37,7 @@ Git URLs used for npm packages can be formatted in the following ways:
 - `git+https://user@hostname/project/blah.git#commit-ish`
 
 The `commit-ish` can be any tag, sha, or branch that can be supplied as
-an argument to `git checkout`. The default `commit-ish` is `master`.
+an argument to `git checkout`. The default `commit-ish` is `main`.
 
 ## About modules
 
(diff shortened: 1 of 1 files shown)


content/packages-and-modules/introduction-to-packages-and-modules/about-packages-and-modules.mdx
```diff
@@ -37,7 +37,7 @@ Git URLs used for npm packages can be formatted in the following ways:
 - `git+https://user@hostname/project/blah.git#commit-ish`
 
 The `commit-ish` can be any tag, sha, or branch that can be supplied as
-an argument to `git checkout`. The default `commit-ish` is `master`.
+an argument to `git checkout`. The default `commit-ish` is `main`.
```
ljharb, line 40: it's a bit more complex than that, since a) repos from before the change could still have `master`, and b) you can set any default branch you want. what does npm actually do here?

wraithgar: There is no default. The default is to not use one and let github.com default it.  When a url requires one (such as with the tarball url) `HEAD` is used.


```javascript
> require('hosted-git-info').fromUrl('github:npm/cli').https()
'git+https://github.com/npm/cli.git'
> require('hosted-git-info').fromUrl('github:npm/cli').tarball()
'https://codeload.github.com/npm/cli/tar.gz/HEAD'
```

seantcoyote:
Thanks @ljharb - I committed your suggestion.

PR #592, opened 2023-04-20
doc: missing npm run build step

peterpeterparker:
I copied the snippet provided to create a GitHub action and used it to publish to npm.
As `npm run build` is missing in the action, I ended publishing an empty package to npmjs.


wraithgar:
That step isn't needed in every project though.

ljharb:
Typically, if the build step is required for publishing, you'd put `npm run build` in a `prepublishOnly` script, so you don't have to remember to do it.

ljharb:
There's also a chance your package has no files in it in the first place - copy-pasted snippets can't work for every possible situation, but if you're following best practices (like including your build process in a prepublish step) then it should work fine.

peterpeterparker:
Let's close it then.

PR #593, opened 2023-04-21
Fix id-token permissions doc in generating-provenance-statements.mdx

pragmaticivan:
Ref: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

The example bellow is correct. GHA docs use the plural.

## References
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs


pragmaticivan:
duplicated

PR #636, opened 2023-05-16
Added clarification about .npmrc and setup-node

iiromutikainen:
Project-specific `.npmrc` file does not work if Github Actions' `setup-node` is used. Though this is more GA related thing, I think it would be good to mention about it in NPM's documentation also.

### DIFF
--- a/content/integrations/integrating-npm-with-external-services/using-private-packages-in-a-ci-cd-workflow.mdx
+++ b/content/integrations/integrating-npm-with-external-services/using-private-packages-in-a-ci-cd-workflow.mdx
@@ -71,6 +71,8 @@ Consult your CI/CD server's documentation for more details.
 
 ## Create and check in a project-specific .npmrc file
 
+**NOTICE:** This **DO NOT** work if you use `setup-node` action in Github Actions because it creates the `.npmrc` file automatically. Instead, see https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs#example-using-a-private-registry-and-creating-the-npmrc-file to configure it correctly.
+
 Use a project-specific `.npmrc` file with a variable for your token to securely authenticate your CI/CD server with npm.
 
 1. In the root directory of your project, create a custom `.npmrc` file with the following contents:


MylesBorins (changes requested):
I don't believe that we should have instructions for specific providers (even GitHub) in these docs

iiromutikainen:
I don't mean explicit instructions for specific providers, but a mention about this specific case since GitHub Actions + _setup-node_ is so popular. There's also an example of GitHub Actions above: "_For example, in GitHub Actions, you would [add your token as a secret](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets). Then you can make the secret available to workflows._"

MylesBorins:
@iiromutikainen Looking [at the source for setup-node](https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/src/authutil.ts#L37-L45) it seems like it appends an existing `.npmrc` if one exists, which is important if multiple scopes or third party registries are being defined. It also looks like it only modifies the `.npmrc` [if you specify a `registry-url`](https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/src/main.ts#L58C1-L60) in the `setup-node`, which is optional. The configuration examples shown in the `setup-node` README don't show this specific pattern.

With this in mind it appears, although I haven't tested it, that if you were to follow the basic pattern as outlined in the docs here it should work, even if you use setup node.

As such I still think that this very specific caveat isn't needed, this makes more sense to be covered in the GitHub actions docs

iiromutikainen:
> @iiromutikainen Looking [at the source for setup-node](https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/src/authutil.ts#L37-L45) it seems like it appends an existing `.npmrc` if one exists, which is important if multiple scopes or third party registries are being defined. It also looks like it only modifies the `.npmrc` [if you specify a `registry-url`](https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/src/main.ts#L58C1-L60) in the `setup-node`, which is optional. The configuration examples shown in the `setup-node` README don't show this specific pattern.
> 
> With this in mind it appears, although I haven't tested it, that if you were to follow the basic pattern as outlined in the docs here it should work, even if you use setup node.
> 
> As such I still think that this very specific caveat isn't needed, this makes more sense to be covered in the GitHub actions docs

Thanks. After taking a bit more detailed look into GA's and `setup-node`, it seems that `registry-url` should be used with private packages: [https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#use-private-packages](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#use-private-packages)

Another option is to add a step to GA workflow with `echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc` which seems to work. What makes this strange is that when I add identical custom `.npmrc` file to the root, it does not work: with custom `.npmrc` file _without_ `setup-node` it throws "_404 xxx is not in this registry._"

Anyway this seems like GA and `setup-node` related thing - closing.

PR #666, opened 2023-06-29
Fixed typo

devjhex:
Fixed typo on line 42
```js
var projectName = require("@scope/package-name") /*No semicolon*/ 

var projectName = require("@scope/package-name");  /*Added semicolon*/ 
```
Fixed typo on line 42 by adding a semi-colon after the variable assignment that would cause a syntax error since it is used to separate different statements in JavaScript.

The request makes changes on the syntax by adding a semi colon after the projectName variable. It makes a syntax fix to the documentation for other users hence not getting some unexpected errors when the code is copied.


wraithgar:
the semicolon is not required and most modern linting rules remove them.

devjhex:
Ok, I just thought it would be a minor fix for those who would just copy and paste the code from the documentation to their code or wherever they're pasting it to.

PR #684, opened 2023-08-08
Add example for publishing with GitLab provenance

remcohaszing:
This adds an example got publishing with provenance from GitLab CI.

Also code blocks in that document were tagged with their language to provide syntax highlighting. Also the inconsistent indentation was fixed for the GitHub workflow example.

## References

Related to https://github.com/npm/feedback/discussions/968

PR #686, opened 2023-08-11
Add GitLab CI to the list of supported CI providers for provenance

travi:
https://github.com/npm/cli/pull/6375 and https://github.com/npm/documentation/pull/684 highlight that GitHub Actions is no longer the only supported provider. we currently link from our semantic-release docs for configuring GitHub Actions, and I would like to do similar for GitLab CI


wraithgar (approved):
Hooray! Thanks.

PR #697, opened 2023-09-03
Update downloading-and-installing-node-js-and-npm.mdx

janat08:
suggest nix pure package manager to manage versions (download link)


wraithgar:
This list is increasingly hard to maintain and we are probably going to pare it down and change it to match what is in [npm's readme](https://github.com/npm/cli#node-version-managers):

> If you're looking to manage multiple versions of Node.js &/or npm, consider using a [node version manager](https://github.com/search?q=node+version+manager+archived%3Afalse&type=repositories&ref=advsearch)

As of now we're not adding to the list.

ME3149235472:
LICENSE #700

PR #698, opened 2023-09-05
Update creating-and-publishing-scoped-public-packages.mdx

ainizoda:
fixed typo


wraithgar:
This is not a typo it's escaping the italicization. This change was made in https://github.com/npm/documentation/pull/133.  This change will just move the backslashes around.

PR #714, opened 2023-09-26
fix: add eslint-plugin-github and fix lint errors

lukekarrys:


theme/src/components/nav-drawer.js
```diff
@@ -71,44 +70,20 @@ function NavDrawer({location}) {
...
-            )}
-          </Details>
-        ) : (
-          <Link key={index} href={item.url} color="inherit" display="block">
-            {item.title}
-          </Link>
-        )}
-      </BorderBox>
-    )
-  })
```
lukekarrys, line 111: Same with this. It was uncalled code, so I removed it.

theme/src/components/code.js
```diff
@@ -28,6 +28,8 @@ function Code({className, children, live, noinline}) {
         {({className, style, tokens, getLineProps, getTokenProps}) => (
           <BorderBox as="pre" className={className} mt={0} mb={3} p={3} border={0} style={{...style, overflow: 'auto'}}>
             {/* This is the scroll handle, it is supposed to be focused with keyboard and scroll a wide codebox horizontally */}
+            {/* FIXME: only adding this line temporarily so it shows up in the PR diff. I'm not sure how to correctly label this semantically. */}
+            {/* eslint-disable-next-line github/a11y-role-supports-aria-props */}
```
lukekarrys, line 32: @github/accessibility-reviewers can I get a review on this PR, specifically this area?

Here is a page on the live site that has some of these code blocks rendered on it: https://docs.npmjs.com/about-scopes

kendallgassner: Would it make sense to put this in a semantic `<code>`

lukekarrys: I think so. If I do that, is it ok to use `visibility: hidden` and remove the `aria-label`? That what I implemented in my latest commit.  

There are no longer any lint warnings and the functionality seems the same to me, but is it still accessible?

lukekarrys:
@kendallgassner as we talked about in office hours, I changed this element back to a `<div aria-hidden="true">` and removed the `aria-label` since it is only being used as an invisible scroll handler.

I was also able to add a `<code></code>` element around the actual code block.

Here is a link to the relevant portion of the diff: https://github.com/npm/documentation/pull/714/files#diff-a937b27cceec7399880913557eacdd40b952463cbbdd2a10ffed757124437a8eR57-R66 Does that look good to you?

theme/src/components/code.js
```diff
@@ -28,44 +54,21 @@ function Code({className, children, live, noinline}) {
...
-            </div>
-            {tokens.map((line, i) => (
-              <div key={i} {...getLineProps({line, key: i})}>
-                {line.map((token, key) => (
-                  <Text key={key} fontFamily="mono" fontSize={1} {...getTokenProps({token, key})} />
-                ))}
-              </div>
-            ))}
+            <div aria-hidden="true" style={{position: 'absolute', ...scrollSize}} ref={scrollHandleRef} />
+            <code>
```
kendallgassner, line 58:
```suggestion
   
```

theme/src/components/code.js
```diff
@@ -28,44 +54,21 @@ function Code({className, children, live, noinline}) {
...
+              {tokens.map((line, i) => (
+                <div key={i} {...getLineProps({line, key: i})}>
+                  {line.map((token, key) => (
+                    <Text key={key} fontFamily="mono" fontSize={1} {...getTokenProps({token, key})} />
+                  ))}
+                </div>
+              ))}
+            </code>
```
kendallgassner, line 66:
```suggestion
```

PR #766, opened 2023-10-16
Bugfix 5931: Scrollable content is not accessible on npm docs pages

soudher1:
Ref: https://github.com/github/accessibility-audits/issues/5931

Scrollable content (HTML \<pre\> elements) are not getting focussed while using keyboard tab key to navigate the pages.

Fix is to add the tabIndex property with value 0 for all those elements.

### DIFF
--- a/theme/src/layout/index.js
+++ b/theme/src/layout/index.js
@@ -1,6 +1,6 @@
 import {BorderBox, Box, Flex, Grid, Heading, Position, StyledOcticon, Text} from '@primer/components'
 import {ChevronDownIcon, ChevronRightIcon} from '@primer/octicons-react'
-import React from 'react'
+import React, {useEffect} from 'react'
 import Head from '../components/head'
 import Header, {HEADER_HEIGHT} from '../components/header'
 import PageFooter from '../components/page-footer'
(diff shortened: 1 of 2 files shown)


nishantms (approved):

theme/src/layout/index.js
```diff
@@ -12,12 +12,17 @@ import VariantSelect from '../components/variant-select'
...
 
 function Layout({children, pageContext, location}) {
   const {title, description, status, source} = pageContext.frontmatter
 
   const variantRoot = NavHierarchy.getVariantRoot(location.pathname)
 
+  useEffect(() => {
+    makePreElementsFocusable()
+  }, [children])
+
```
lukekarrys, line 25: The `<pre>` elements are rendered from `theme/src/components/code.js`, so I think we can add a `tabIndex={0}` property to those elements instead of doing it asynchronously in a `useEffect`.

lukekarrys (changes requested):

soudher1:
PR https://github.com/npm/documentation/pull/771 replaces this pr. I will address the review comment in the new pr.
Using branch on master instead of forking the repo

PR #771, opened 2023-10-16
Bugfix 5931 scrollable content accessibility

soudher1:
Ref: https://github.com/github/accessibility-audits/issues/5931

Scrollable content (HTML \<pre\> elements) are not getting focussed while using keyboard tab key to navigate the pages.

Fix is to add the tabIndex property with value 0 for all those elements.


nishantms (approved):

soudher1:
.deploy production

PR #774, opened 2023-10-17
Fixed Signin color contras issue

dipeshpatel01:
Fixed The link-'Sign In' Contras ratio and underline.

https://github.com/github/accessibility-audits/issues/6060

Screenshot 2023-10-20 at 11 43 17 AM

### DIFF
--- a/src/shared.js
+++ b/src/shared.js
@@ -6,7 +6,7 @@ import Note from '../theme/src/components/note'
 export default {
   /* User login */
   'user-login': {
-    text: (<>On the npm "<Link href="https://www.npmjs.com/login" underline={true} color="#0071F1">Sign In</Link>" page, enter your account details and click <strong>Sign In</strong>.</>),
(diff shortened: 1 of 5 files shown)


lukekarrys (changes requested):
This is likely an issue with all the links on the site since we use the same link color and background color for them.

To change all the links we should update the styles on the `<a>` component in `theme/src/wrap/root-element.js`.

dipeshpatel01:
> This is likely an issue with all the links on the site since we use the same link color and background color for them.
> 
> To change all the links we should update the styles on the `<a>` component in `theme/src/wrap/root-element.js`.

My guess is If we apply the in root element. all the links including sidebar also will affect the result of underline and color. so we should apply style in role="region" id="skip-nav" component in `theme/src/layout/index.js

lukekarrys:
This was fixed by #778.

PR #775, opened 2023-10-18
[5940] and [5936] show table of contents when browser window gets small

david-edwards01:
https://github.com/github/accessibility-audits/issues/5940
https://github.com/github/accessibility-audits/issues/5936

show table of contents when browser window gets small

closes #5940 and #5936

Same fix for both. Yes it is reproducible that the controls under Table of Contents is not visible in a small screen (only the title is visible "Table of Contents" - and it has to be clicked to see the list. The mouse does not indicate that the title is clickable). I made the the entire Table of Contents block visible (not collapsed) as it seems to be asking for. If not, it just needs the "Pointer" hand cursor for "Table of Contents" (maybe tabindex="0").

UnderTOC

### DIFF
--- a/theme/src/layout/index.js
+++ b/theme/src/layout/index.js
@@ -1,5 +1,4 @@
-import {BorderBox, Box, Flex, Grid, Heading, Position, StyledOcticon, Text} from '@primer/components'
-import {ChevronDownIcon, ChevronRightIcon} from '@primer/octicons-react'
+import {BorderBox, Box, Flex, Grid, Heading, Position, Text} from '@primer/components'
 import React from 'react'
 import Head from '../components/head'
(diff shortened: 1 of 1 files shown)


david-edwards01:
image

Table of Contents remains visible on smaller screen by stacking it with other content and maintaining the same tab cursor order as it had when it was on the right column in a larger window.

david-edwards01:
Error: Ensure GITHUB_TOKEN has permission "id-token: write".
token-write-error

@nishantms
@lukekarrys  
@wraithgar 
@alexandra-tashjian

lukekarrys (changes requested):
This will be fixed by #792.

PR #767, opened 2023-10-16
Migrate to @primer/react

lukekarrys:
This also merges the `theme/` directory into the root since there is not a reason to keep it separate now that it is all in one repo. This removes the file shadowing that is possible with Gatsby themes, as well as us needing to support passing options to the theme. This PR also attempt to remove any dead code found during this migration.

- [x] Remove prettier from cli workspace
- [x] Add content to 404 page
- [x] Pages with code blocks dont resize properly
- [x] Change sidebar spacing/text size
- [x] code blocks have extra new line
- [x] code blocks are not syntax highlighted

<details><summary>Code Block Screenshot</summary>
<p>

image

</p>
</details> 

<details><summary>Resize Screenshot</summary>
<p>
Screenshot 2023-10-17 at 10 50 58 AM
</p>
</details> 

Closes #765 
Closes #764 
Closes #738 
Closes #734 
Closes #772 
Closes #770
Closes #760

PR #777, opened 2023-10-18
[5940] TOC has cursor pointer - clickable dropdown

david-edwards01:
When window gets small, "Table of Contents" becomes a closed dropdown. This fix adds the cursor pointer to "Table of Contents" so that the mouse hover shows it as clickable.

### DIFF
--- a/theme/src/layout/index.js
+++ b/theme/src/layout/index.js
@@ -91,7 +91,7 @@ function Layout({children, pageContext, location}) {
                   <Details>
                     {({open}) => (
                       <>
-                        <Text as="summary" fontWeight="bold">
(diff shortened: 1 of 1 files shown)


david-edwards01:
Choose PR 776 for open TOC, Choose PR 777 for closed dropdown with point hand cursor
[https://github.com/github/accessibility-audits/issues/5940](url)
and duplicate [https://github.com/github/accessibility-audits/issues/5936](url)
Choose_TOC_open_or_closed

lukekarrys:
With #767 the table of contents `<summary>` as been changed to `<Button as="summary">` so it now has a cursor pointer by default. So this is no longer needed.

PR #828, opened 2023-11-02
Update resolving-eacces-permissions-errors-when-installing-packages-g…

yudizaxay:
Instead of ```source ~/.profile``` we have to use ```source ~/.npm-global```


lukekarrys:
In the docs sourcing `~/.profile` is necessary after adding the `PATH` environment variable to it. Sourcing `~/.npm-global` will not work for that.

PR #832, opened 2023-11-03
docs: add fnm to downloading-and-installing-node-js-and-npm.mdx

pure-js:
Why not to add the fast node version manager to the docs?


wraithgar:
This isn't really a list we are wanting to have to curate and maintain.

In the cli we changed this to a [github search](https://github.com/npm/cli/blob/e91d5c64a99ad5c977d292a3d3383f4cfc0d1158/docs/lib/content/configuring-npm/install.md?plain=1#L36)

If this file gets updated at all it will be to fall in line w/ that approach.

PR #889, opened 2023-12-19
doc: update npm-config link in npm-edit descriptions

lsst25:
Fixed some broken links in the 'npm-edit' command docs (v6, v7, v8, v9, v10). They pointed to 'npm-config' before but now they're heading to the right place, 'commands/npm-config'. 

Your review is appreciated.


wraithgar:
All of the files in `content` are sourced from the npm repo.  You'll have to update it there, and we'll likely only update the latest one (v10).  Older docs are essentially frozen.

https://github.com/npm/cli/blob/latest/docs/lib/content/commands/npm-edit.md

PR #974, opened 2024-02-13
docs: Add conventional commit note to contributing.md

wraithgar:
Copied from the current version in template-oss, and adapted to fit the docs site instead of a published module.


wraithgar:
This copy may not be appropriate here since we don't publish this as a package.

dhei (approved):

PR #1135, opened 2024-06-03
Add example of undeprecating entire package

karlhorky:
Hi npm team, first of all, thanks for your continued effort on npm and documentation!

Add a command example of undeprecating an entire package, which appears to be missing from the documentation currently.

The closest thing currently is this heading + sentence:

> # Undeprecating a package or version
>
> To undeprecate a package, replace "<message>" with "" (an empty string) in one of the above commands.

But I would argue it's a bit open to interpretation, and doesn't explicitly show the command to undeprecate an entire package.

## References

--


wraithgar (approved):

karlhorky:
@wraithgar thanks for the review and merge!

PR #1152, opened 2024-06-12
chore: upgrade

reggi:
the issue with dependabot is that [eslint-plugin-primer-react](https://github.com/primer/eslint-plugin-primer-react) and eslint-plugin-github need to be bumped at the same time. This PR bumps both of them to the latest version and upgrades everything else too.


wraithgar:
That lockfile change seems suspicious.

wraithgar (approved):

PR #1177, opened 2024-06-26
docs: how to run on windows locally

### LINKED ISSUE #1172: Using cross-env for cross platform compatibility
I ran into this same issue: https://github.com/npm/documentation/issues/861 - the suggestion to use the set command for windows works fine. However, can we not use `cross-env` to make the npm commands work seamlessly across operating systems? I would like to make this change if that is a good idea. The changes would be in `package.json` file:
* A dependency on `cross-env` package.
* Prefixing `gatsby` and `develope:ssr` scripts with `cross-env` as they are the only two commands with environment variables.

If using `cross-env` is not a good idea, then perhaps we could mention in the `CONTRIBUTING.md`  guide to to use windows specific syntax where necessary.

andrewnessinjim:
The syntax for setting env variables is different for powershell and command prompt, so I have mentioned only the command prompt syntax to keep it simple. I think it's sufficient if Windows users just know the syntax needs to be updated.

We could add the windows syntax for setting `GATSBY_CONTENT_ALLOW` and `GATSBY_CONTENT_IGNORE` too, but it feels like an overkill.

Tested by changing the syntax in `package.json` file:
image

Closes #1172

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ This is the documentation for [https://docs.npmjs.com/](https://docs.npmjs.com/)
 
 First, `npm install` the dependencies. This will install Gatsby, et al.
 
-Next, `npm run develop` to start the test server to view your changes. The Gatsby server will be started on port 8000. You can navigate to `http://localhost:8000` to view the site live.
+Next, `npm run develop` to start the test server to view your changes. The Gatsby server will be started on port 8000. You can navigate to `http://localhost:8000` to view the site live. If you're on Windows, change the syntax for setting environment variables in the `package.json` file based on your terminal before running this command. The syntax for command prompt is `set ENV_VAR=value & command`.
 
 In order to cut down on the time it takes `npm run develop` to complete, you can use the environment variables `GATSBY_CONTENT_ALLOW` and `GATSBY_CONTENT_IGNORE` to only build some pages.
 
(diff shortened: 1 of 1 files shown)


andrewnessinjim:
I updated the commit message to have "docs" prefix instead of "feat" and forced pushed the commit.

CONTRIBUTING.md
```diff
@@ -25,7 +25,7 @@ This is the documentation for [https://docs.npmjs.com/](https://docs.npmjs.com/)
 
 First, `npm install` the dependencies. This will install Gatsby, et al.
 
-Next, `npm run develop` to start the test server to view your changes. The Gatsby server will be started on port 8000. You can navigate to `http://localhost:8000` to view the site live.
+Next, `npm run develop` to start the test server to view your changes. The Gatsby server will be started on port 8000. You can navigate to `http://localhost:8000` to view the site live. If you're on Windows, change the syntax for setting environment variables in the `package.json` file based on your terminal before running this command. The syntax for command prompt is `set ENV_VAR=value & command`.
```
wraithgar, line 28: If this file gets changed, folks are almost certainly going to accidentally commit those changes, which we don't want.

We may just have to say that local development on windows doesn't work with the shortcut scripts, and that you have to manually run the associated commands.

andrewnessinjim: Makes sense! Should I amend the existing commit or add a new commit?

wraithgar: You can keep adding commits, we can squash the PR when it's ready.

andrewnessinjim:
I tested the command:
image

wraithgar (approved):

PR #1195, opened 2024-07-16
feat: move ci to node 22

wraithgar:
Some deps are starting to drop node 18. Might as well go to the latest


reggi:
I think you forgot `.nvmrc` if intentional no worries!

wraithgar:
Not intentional, that's outside of the domain of template-oss so it was forgotten, showing why template-oss is so important lol.

Updated `.nvmrc` and `CONTRIBUTING.md`

PR #1283, opened 2024-09-17
adds devEngines

reggi:
This adds documentation for the devEngines feat.

https://github.com/npm/cli/pull/7766


wraithgar:
Looks good so far.  We'll probably want to wait till the cli PR lands so we are sure of when it actually applies.

reggi:
this needed to be added in the cli because `package.json` doc is pulled in from there

PR #1319, opened 2024-10-10
Accessibility#9196

maitxn:
## What/Why

- Screen reader does not announce the search result information as `No results` on providing invalid input

## Changes

- [`src/components/search.js`](https://github.com/npm/documentation/pull/1319/files#diff-995f95946b0cbf613b5cf5d8d6335cb1612dcb31907922cd5da28be90a6ff442)

> Add `aria-live` as `polite` to Box element

> Add `announce` function to trigger screen reader announcing the `No results` message when there is no search result

## Before
image

## After

- The screen reader announces the search result information as `No results` on providing invalid input

## References

- https://github.com/github/accessibility-audits/issues/9196

### DIFF
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -14,16 +14,10 @@ import {announce} from '../util/aria-live'
 
 const SearchResults = ({results, getItemProps, highlightedIndex, isOpen}) => {
   const siteMetadata = useSiteMetadata()
-
-  React.useEffect(() => {
-    if (isOpen && (!results || results.length === 0)) {
-      announce('No results')
-    }
-  }, [results, isOpen])
-
   if (!results || results.length === 0) {
+    announce('No results')
     return (
-      <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="assertive">
+      <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="polite">
         No results
       </Box>
     )
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -22,7 +22,11 @@ const SearchResults = ({results, getItemProps, highlightedIndex, isOpen}) => {
   }, [results, isOpen])
 
   if (!results || results.length === 0) {
-    return <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="assertive">No results</Box>
+    return (
+      <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="assertive">
+        No results
+      </Box>
+    )
   }
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/search.js
```diff
@@ -10,12 +10,23 @@ import {LightTheme} from '../theme'
...
+    }
+  }, [results, isOpen])
+
   if (!results || results.length === 0) {
-    return <Box sx={{fontSize: 2, px: 3, py: 3}}>No results</Box>
+    return (
+      <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="assertive">
+        No results
+      </Box>
+    )
```
leobalter, line 29:
```suggestion
    announce('No Results')
    return (
      <Box sx={{fontSize: 2, px: 3, py: 3}} aria-live="polite">
        No Results
      </Box>
    )
```

leobalter: I'm not sure the useEffect is needed, but the announce seems to work the same way with the announce in here instead.

There is one caveat: this code is still failing to _announce_ the "No Results" on the first No Results search, but passing on the consecutive tries.

maitxn: I tried to use useEffect to resolve the first No result search but it doesn't seem to work this way. The issue is fixed as your suggested change but there is still one caveat failing on the first No result search

leobalter (approved):
> I tried to use useEffect to resolve the first No result search but it doesn't seem to work this way. The issue is fixed as your suggested change but there is still one caveat failing on the first No result search

the code looks better and indeed the first No result is still ongoing but for Chromevox only.

Did you have a chance to try this with a different screenreader other than chromevox? 

It works ok using VoiceOver (cmd+F5 on MacOS). I didn't test this on JAWS, NVDA, etc but I take the implementation seems correct (and Chromevox is the one failing).

maitxn:
> > I tried to use useEffect to resolve the first No result search but it doesn't seem to work this way. The issue is fixed as your suggested change but there is still one caveat failing on the first No result search
> 
> the code looks better and indeed the first No result is still ongoing but for Chromevox only.
> 
> Did you have a chance to try this with a different screenreader other than chromevox?
> 
> It works ok using VoiceOver (cmd+F5 on MacOS). I didn't test this on JAWS, NVDA, etc but I take the implementation seems correct (and Chromevox is the one failing).

I don't have JAWS, NVDA. It is working ok using VoiceOver on Mac

maitxn:
@wraithgar FYI this PR got approved from @leobalter. Can you please merge it to prod?

PR #1402, opened 2024-12-05
feat: preparation for v11 & gh api rate limiting

reggi:
Adds a release separation for v10 and v11 branches in npm.

* adds `gh` exec to get token so can be run locally
* adds caching for each version sha, and if match dont fetch files `cache.json`
* drops fetching v6 / v7 branches
* pulls in 11

A part of npm 11 roadmap: https://github.com/npm/statusboard/issues/898

PR #1417, opened 2024-12-18
fix: remove v11 changes from v10 changelog

### LINKED ISSUE #1416: Discrepancy between v10 vs v11 changelog
v10:
image

v11:
image

Note the additional breaking changes listed under the v11 releases when viewing the v11 changelog in comparison to the v10 changelog.

Having said that... shouldn't the v11 changes be hidden when viewing the v10 changelog?

JoelGritter:
The inclusion of the v11 changelog entries in the v10 changelog appears to be accidental.

Additionally, it is causing a discrepancy between the breaking changes listed for v11. Fewer breaking changes were listed on the v10 changelog for the v11 changes compared to the v11 changelog.

## References
Fixes #1416


wraithgar:
These files are auto generated from the cli's actual source code.  Changes here will be overwritten every time the docs site redeploys/rebuilds.

It looks possibly like when https://github.com/npm/documentation/pull/1402 was made, the local copy of the cli didn't have the changes in https://github.com/npm/cli/pull/7943 pulled yet.

We'll have to look into this, but this PR is not the way to fix it.

PR #1427, opened 2024-12-27
Add Missing "contributors" and "maintainers" info in v10/package.json `people` field

kenshanta:
Loosely based on [issue #113](https://github.com/npm/www/issues/133)


wraithgar:
This file is generated from https://github.com/npm/cli/blob/latest/docs/lib/content/configuring-npm/package-json.md.

If we want to update this we need to be extra clear on what parts the user adds, and what parts are added by the npm registry.  Not every registry adds those fields.

PR #1443, opened 2025-01-15
Update disputes.mdx to clarify squatting dispute process

thomasmost:
The previous phrasing is extremely confusing; does this policy apply or does it not apply? It sounds as if those documents refer to this one but this one does not apply.

Adding an 'except' I believe produces the expected framing...

...but truly who knows! I still haven't received an update on [my squatting dispute](https://support.github.com/ticket/personal/0/2910722) and it's been over 5 months


thomasmost:
@leobalter could use some clarity on this

Issue #1357, opened 2024-11-04
Add Dark Mode to npm Documentation Site UI


DENNIS-CODES:
#### Summary
Many developers prefer dark mode for better comfort and reduced eye strain during extended use. This issue proposes the addition of a dark mode feature to the npm documentation site, providing a more developer-friendly interface. This enhancement would align with current industry standards, as dark mode is now widely implemented across developer tools and documentation platforms.

#### Motivation
As developers, we often work late hours and prefer dark themes to minimize eye strain and avoid disrupting our circadian rhythms. With a substantial number of developers using npm regularly, adding dark mode to the npm documentation website would enhance usability and make the site more accessible.

**Reasons for Implementing Dark Mode:**
- **Reduced Eye Strain**: Dark mode reduces the intensity of light entering the eyes, which helps in long coding sessions.
- **Battery Efficiency**: On OLED screens, dark backgrounds consume less power, making it a more battery-friendly option.
- **Aesthetic Consistency**: Many developer tools and environments (e.g., VSCode, GitHub) already offer dark mode, allowing a seamless experience for developers switching between tools.

#### Expected Outcome
The npm documentation site should provide users with the option to switch between light and dark modes. Ideally, the dark mode would:
- Be accessible via a toggle button in the UI
- Respect user preferences by defaulting to dark mode if the user has enabled dark mode in their operating system or browser settings
- Be designed with accessibility in mind, ensuring sufficient contrast for readability

#### Proposed Solution
1. **CSS Variables**: Define a set of CSS variables (e.g., `--bg-color`, `--text-color`) that can be adjusted based on the selected theme.
2. **Theme Toggle**: Add a theme toggle switch in a prominent location (e.g., the top-right corner of the page).
3. **Media Queries for System Preferences**: Use CSS media queries (e.g., `@media (prefers-color-scheme: dark)`) to automatically enable dark mode based on the user’s system preference.

#### Example

Below is a quick mock-up to illustrate the dark mode implementation. CSS variables would adjust the color scheme across the entire documentation site:

```css
/* Define color variables for light and dark mode */
:root {
  --bg-color-light: #ffffff;
  --text-color-light: #000000;
  --bg-color-dark: #121212;
  --text-color-dark: #e0e0e0;
}

body {
  background-color: var(--bg-color-light);
  color: var(--text-color-light);
}

body.dark-mode {
  background-color: var(--bg-color-dark);
  color: var(--text-color-dark);
}

/* Automatically detect user preference */
[...]

PR #1480, opened 2025-02-13
Feature: Dark mode

### LINKED ISSUE #839: dark mode
can add dark mode?

laine-hallot:
## Info
This PR adds a way to change the site's color mode between light and dark. The features been requested a couple times before and it just seemed like a nice thing to have.

## Dev Notes
I just had it use the default `dark` theme from `@primer/react` for dark/night mode since it seemed like the project more or less uses the default light theme save for a red accent color. I did try using that accent color in dark mode but I felt like the contrast just wasn't good enough so I left it as the default blue accent color.

This PR also adds a couple hooks deal with switching the color mode.
#### `useColorPreference()`
This hook handles changing the theme context's `colorMode` as well as setting the selected value in localStorage.

Since the value in localStorage and the theme context are separate and there's no guarantee that every call of the hook is happening in the same theme context the hook accepts a parameter to change the prefix for key in local storage which at least gives you a way to call `useColorPreference` in multiple theme context without them overwriting each other's  localStorage values.
You can still overwrite one context's localStorage value with another's if you call `useColorPrefrence` using same prefix value. 
  
#### `usePrismTheme()` 
This one just checks what the effective color mode is for the current theme context and returns a Prism theme. I set it to use `vsDark` for dark/night mode since that seems like the default for Prism.

## Screenshots

Color demo
npm-doc-dark-mode-demo-desktop

Code theme demo: 
npm-doc-dark-mode-code-demo

Persistence: 
npm-doc-dark-mode-persistance-demo

Mobile:
npm-doc-dark-mode-demo-mobile

## References
Closes #839

### DIFF
--- a/src/components/header.js
+++ b/src/components/header.js
@@ -55,7 +55,7 @@ function Header() {
               </Link>
             ))}
           </Box>
-          <Box sx={{display: ['flex', null, null, 'none']}}>
+          <Box sx={{display: ['flex', null, null, 'none'], ml: [4, null, null, 'unset']}}>
             <Search.Mobile {...search} />
             <NavDrawer />
           </Box>


laine-hallot:
@leobalter sorry to ping you, this PR has just been open for a few months and I wasn't sure if anyone even saw it at this point.
Is this a feature the maintainers are interested in adding at all? I know you said the feature wasn't a priority here (https://github.com/npm/documentation/issues/1357#issuecomment-2504592046) but I figure that's mostly because that was in response to a feature request and assumed the code still needed to be written by one of the core contributors instead of just a PR that implemented it.

PR #1550, opened 2025-04-13
docs: clarify token creation in npm login command

LiangYingC:
This is simple PR for adjust `npm login` doc.

### Problem Description:

When using `npm login`, users are not aware that this command generates a legacy token with publish permissions. This information is not evident in the `npm login` cli documentation.

After a while searching. found that this context and detail hiding in the [About legacy tokens](https://docs.npmjs.com/about-access-tokens#about-legacy-tokens) documentation. 

### Solution:
Add a clear description in the `npm login` documentation in this PR so that users can understand publish token would be created after `npm login` in the future.

### DIFF
--- a/content/cli/v11/commands/npm-login.mdx
+++ b/content/cli/v11/commands/npm-login.mdx
@@ -47,6 +47,8 @@ Note: This command is unaware of workspaces.
 
 Verify a user in the specified registry, and save the credentials to the `.npmrc` file. If no registry is specified, the default registry will be used (see [`config`](/cli/v11/using-npm/config)).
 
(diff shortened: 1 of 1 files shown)


wraithgar:
This is a great addition, however that file is pulled externally from the cli repo and needs to be changed [there](https://github.com/npm/cli/blob/latest/docs/lib/content/commands/npm-login.md), not here.  We may also want to add a similar clause to [adduser](https://github.com/npm/cli/blob/latest/docs/lib/content/commands/npm-adduser.md)

LiangYingC:
@wraithgar  thanks for your reminders. I've opened another PR to make this addition. 

https://github.com/npm/cli/pull/8226

PR #1578, opened 2025-05-10
fix: v11 npm run page not found

kennethlieyanto:
Page for "npm run" in v11 not found.
https://docs.npmjs.com/cli/v11/commands/npm-run not found, judging from content/nav.yml all previous version points to /npm-run-script but for some reason the v11 version points to /npm-run.


hashtagchris:
Related: https://github.com/npm/cli/pull/8265

https://docs.npmjs.com/cli/v11/commands/npm-run-script/ returns 200 OK ✅ . However ultimately I think we want to get https://docs.npmjs.com/cli/v11/commands/npm-run/ returning a page as well, since `run` is now the primary command name.

PR #1655, opened 2025-07-01
Update about-semantic-versioning.mdx to clarify version resolution and pre-releases

LBaquedanoCordova:
This pull request adds a note to the "About semantic versioning" guide to clarify how npm resolves versions within defined ranges in `package.json`. It specifies that npm installs the highest compatible **stable** version and ignores pre-releases like `1.2.0-beta.1` unless explicitly allowed by configuration.

The goal is to help developers better understand version behavior and prevent unexpected installations when using semver constraints.

## References
None

### DIFF
--- a/content/packages-and-modules/contributing-packages-to-the-registry/about-semantic-versioning.mdx
+++ b/content/packages-and-modules/contributing-packages-to-the-registry/about-semantic-versioning.mdx
@@ -35,6 +35,12 @@ For example, to specify acceptable version ranges up to 1.0.4, use the following
 
 For more information on semantic versioning syntax, see the [npm semver calculator][semver-calc].
 
+<Note>
+
+**Note:** When using version ranges in your `package.json`, npm resolves to the **highest available version** that satisfies the specified constraints. This behavior ensures that consumers automatically receive the most recent compatible stable release, as long as the changes remain within the defined semver range. Pre-release versions (e.g., `1.2.0-beta.1`) are ignored unless explicitly requested or allowed via specific configuration.
+
+</Note>
+
 ### Example
 
 ```json


content/packages-and-modules/contributing-packages-to-the-registry/about-semantic-versioning.mdx
```diff
@@ -35,6 +35,12 @@ For example, to specify acceptable version ranges up to 1.0.4, use the following
 
 For more information on semantic versioning syntax, see the [npm semver calculator][semver-calc].
 
+<Note>
+
+**Note:** When using version ranges in your `package.json`, npm resolves to the **highest available version** that satisfies the specified constraints. This behavior ensures that consumers automatically receive the most recent compatible stable release, as long as the changes remain within the defined semver range. Pre-release versions (e.g., `1.2.0-beta.1`) are ignored unless explicitly requested or allowed via specific configuration.
+
```
wraithgar, line 41: I believe that if your range includes a prerelease then new prereleases will be included.
example from semver.npmjs.com showing >11.0.0-pre.0 including 11.0.0-pre.1

LBaquedanoCordova: Thanks for your input! You're absolutely right that when the version range explicitly includes a prerelease version — like `>11.0.0-pre.0` — then further prereleases will be considered as valid candidates.

My contribution was intended to reflect the default behavior when the version range does **not** explicitly include a prerelease version — for example, using `^1.0.0` or `~2.2.0`. In those cases, even if a prerelease version exists within the semantic range (like `1.1.0-beta.0`), `npm install` will not install it unless:
- the prerelease is explicitly included in the version range, or
- the version is installed directly via `npm install some-lib@1.1.0-beta.0`.

This behavior has even led to some confusion in the community, as discussed in [npm/cli#7851](https://github.com/npm/cli/issues/7851), where users expected `>x.y.z-beta` to include later prereleases — but npm does not currently resolve those by default.

This distinction between stable and prerelease resolution isn't always obvious, so I thought it would be helpful to clarify. Happy to refine the wording further if needed!

PR #1681, opened 2025-07-23
Update creating-an-organization.mdx

Memohalliwell:
Just change one word i think sounds better your organization


owlstronaut:
Sorry, I think we will keep the wording as-is

PR #1710, opened 2025-09-10
fix: gitlab example for trusted publishing

jrvidal:
I haven't been quite able to make trusted publishing work in GitLab (see https://github.com/npm/cli/issues/8558), but I do think the example is incorrect. `id_tokens` is not a valid global keyword. It needs to be specified under `default` or a job. I think a job makes more sense.

Reference:
* https://docs.gitlab.com/ci/yaml/#global-keywords
* https://docs.gitlab.com/ci/yaml/#default
* https://docs.gitlab.com/ci/yaml/#id_tokens

### DIFF
--- a/content/packages-and-modules/securing-your-code/trusted-publishers.mdx
+++ b/content/packages-and-modules/securing-your-code/trusted-publishers.mdx
@@ -118,10 +118,6 @@ stages:
 variables:
   NODE_VERSION: '20'
 
-id_tokens:
-  NPM_ID_TOKEN:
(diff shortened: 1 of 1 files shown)


jrvidal:
Yes:

* [Failure](https://gitlab.com/_rvidal/codemirror-lang-scheme/-/pipelines/2032818620)
* [Success](https://gitlab.com/_rvidal/codemirror-lang-scheme/-/pipelines/2032811756)

(The part I haven't made work yet is provenance generation)

PR #1715, opened 2025-09-17
fix: add missing color mapping for token string in colorMap

jpg619:


wraithgar:
Not sure the PR title is appropriate here, this is not a refactor.

owlstronaut (approved):

PR #1760, opened 2025-10-29
Clarify placeholder usage (‘foo’/‘bar’) in CLI flag documentation

SuperHeavyBallet:
The documentation for command-line flags uses “foo” and “bar” as placeholders without clarifying that they are generic examples of a key and a value.

New readers unfamiliar with this convention may interpret them as literal parameters. Suggest adding a short note as given explaining that “foo/bar” are placeholders for any config key/value.

Update:
This branch includes both a minor clarification and a full rewrite. I’m happy to separate them into two PRs if preferred.

### DIFF
--- a/content/cli/v11/using-npm/config.mdx
+++ b/content/cli/v11/using-npm/config.mdx
@@ -25,20 +25,21 @@ npm gets its configuration values from the following sources, sorted by priority
 
 #### Command Line Flags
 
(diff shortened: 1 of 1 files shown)


wraithgar:
We typically do not backport docs changes like this so the older npm versions will stay as-is.

SuperHeavyBallet:
Closing this PR. The change has been migrated to [npm/cli#8708](https://github.com/npm/cli/pull/8708), which is the correct source for this content.

PR #1801, opened 2025-12-04
Add `sudo` to npm upgrade command in example

tstirrat15:
## Description
The example doesn't work as written. According to this issue, it's because you need to use `sudo`: https://github.com/actions/setup-node/issues/324#issuecomment-970306807

This implements that.

## Changes
* Add `sudo` to npm upgrade command

## Testing
Review. See that you can run this workflow successfully.


tstirrat15:
Right. I'm guessing that if you updated that step to point at the most
recent version of `actions/setup-node`, it would start failing.

tstirrat15:
I'm also not asserting that this is an issue with npm, but it is an issue
with the documentation for this specific use case.

PR #1843, opened 2026-01-12
chore: convert box to native components, transform ignore patterns, remove themeget, remove octicon

owlstronaut:


Frankin45:
When will it be ready

PR #1932, opened 2026-04-06
[Design]: Update Search Bar Styling

### LINKED ISSUE #1832: "Search npm Docs" is virtually impossible to distinguish from links in the header bar
See this screenshot:

Image

The search box has _zero_ distinguishing visual features to help people find it. It is so hard to notice this text that I almost concluded (with great irritation) that you simply could not search the npm docs. The problem is that I expected a normal search box control and so when I was visually skimming the header bar, I skipped over the "Search npm Docs" text completely. I'm not going to actually read all of the links in the header bar; I'm going to look specifically for the control that I expect to be there, that looks a particular way in every other website or app.

This problem is made worse by the fact that the links in the header bar don't look very distinctive to begin with: they're not blue, and they're not even underlined. They look like completely normal text and the only clue that they might be clickable is that they're in the header bar. This is not very good UX, and it further serves to make the search box indistinguishable.

Putting a box around the search bar would help. Putting a magnifying glass icon next to the text would also help.

kenshanta:
This PR updates the desktop search bar for better accessibility and visual distinction 

## References
 Fixes #1832 

Screenshot 2026-04-06 at 2 57 03 PM

### DIFF
--- a/src/components/header.module.css
+++ b/src/components/header.module.css
@@ -52,6 +52,9 @@
 .searchDesktop {
   display: none;
   margin-left: 24px;
+  padding-right: 20px;
+  border: 1px solid rgba(255, 255, 255, 0.6);
+  border-radius: 25px;
 }
 
 @media (min-width: 1012px) {
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -81,7 +81,10 @@ export const Desktop = props => {
             <SearchResults {...rest} />
           </LightTheme>
         ) : null}
+
       </div>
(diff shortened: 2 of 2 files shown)


kenshanta:
Recently #1974 got approved and merged.
What am I doing wrong here that I get no feedback of any sort?

owlstronaut:
@kenshanta Sorry, you did nothing wrong. It has more to do with the manpower we have to review all prs across npm. If you want to update this for fine-tune what is on the latest, I'm happy to take a look

PR #1938, opened 2026-04-13
feat: improve search relevance by indexing headings and removing location penalty

karenjli:
### Summary
Improves search by indexing page headings and removing the Fuse.js location-based scoring penalty.

### Changes
- Extract headings from tableOfContents and add as a searchable headings field
- Set ignoreLocation: true so matches anywhere in a document are scored equally
- Add unit tests for the new flattenHeadings helper

### Why
By default, Fuse.js penalizes matches far from position 0. For long body fields, this makes most content effectively unsearchable. Even exact matches deep in a page can get filtered out. Adding headings as a searchable field further improves discoverability of section-level content.

### DIFF
--- a/src/hooks/__tests__/use-search.test.js
+++ b/src/hooks/__tests__/use-search.test.js
@@ -1,0 +1,47 @@
+import {flattenHeadings} from '../use-search'
+
+describe('flattenHeadings', () => {
+  test('returns empty array for null/undefined input', () => {
+    expect(flattenHeadings(null)).toEqual([])
+    expect(flattenHeadings(undefined)).toEqual([])
+  })
+
+  test('returns empty array for empty items', () => {
+    expect(flattenHeadings([])).toEqual([])
+  })
+
+  test('flattens single level of headings', () => {
+    const items = [{title: 'Description'}, {title: 'Configuration'}]
+    expect(flattenHeadings(items)).toEqual(['Description', 'Configuration'])
+  })
+
+  test('flattens nested headings', () => {
+    const items = [
+      {
+        title: 'Description',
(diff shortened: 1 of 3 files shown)


src/hooks/__tests__/use-search.test.js
```diff
@@ -0,0 +1,47 @@
+import {flattenHeadings} from '../use-search'
```
wraithgar, line 1: This is a wholly new test file with a naming 
pattern that doesn't appear to exist in this repo.  Are we sure this is where we want new tests for this?

karenjli: This follows the same pattern as [page-footer.test.js](https://github.com/npm/documentation/blob/main/src/components/__tests__/page-footer.test.js), which is the existing convention for tests under [src](https://github.com/npm/documentation/tree/main/src). Happy to move it if there's a preferred alternative.

wraithgar: Ah yeah I didn't see that one.  *That* seems like a one-off in and of itself.

Idk if we really even have a pattern for testing these files. It appears page-footer.js is the outlier.


Is this the right way to test? Is this testing reality or is it just testing how the code is written?  Do these tests translate to real UX?

wraithgar (approved):

wraithgar:
The test file isn't parsing in jest:

```
Details:

    /home/runner/work/documentation/documentation/src/hooks/use-search.js:182
        worker.current = new Worker(new URL('../util/search.worker.js', import.meta.url));
                                                                               ^^^^

    SyntaxError: Cannot use 'import.meta' outside a module

    > 1 | import {flattenHeadings} from '../use-search'
```

PR #1939, opened 2026-04-14
chore: updateNpm: false

wraithgar:


owlstronaut (approved):

wraithgar:
CLI content failure is cause the build fails which this PR will fix.

PR #1975, opened 2026-05-29
Clarify 2FA docs

nikolas:


owlstronaut (changes requested):
Thanks for looking to clarify some confusion here. The "physical key is required" framing is a little broad, webauthn also covers passkeys. Also stylistically it might fit better as a <Note> block, and we probably don't want to cite community discussions as the authority in the docs

PR #2036, opened 2026-09-10
feat(tokens): document stage-only access tokens

Tayvon:
## What

Adds public documentation for the new **"Read and write (stage only)"** granular access token publish policy.

Part of github/npm#15530 (Phase 1 — safer stage-capable GAT). Closes github/npm#15610.

## Changes

**`about-access-tokens.mdx`** — new `### About stage-only tokens` section covering:
- What it is / why it exists (safer automation on-ramp).
- Capabilities **and** limitations: **can** stage (`npm stage publish`), deprecate, move dist-tags, unpublish; **cannot** directly publish a new version (fails with `E_STAGE_REQUIRED`).
- Promotion flow: `npm stage publish` → maintainer with 2FA runs `npm stage approve <stage-id> --otp <code>`.
- Jan 2027 direct-publish deprecation.
- A `<Note>` explicitly stating this is **not** a general-purpose security boundary (avoids overstating security, per the issue).

**`creating-and-viewing-access-tokens.mdx`** — updates the now-live **Packages and scopes** permission options to match the production UI (wubwub #3762, deployed at 100%): **Read and write (publish and stage)** / **Read and write (stage only)**.

### DIFF
--- a/content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
+++ b/content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
@@ -35,7 +35,7 @@ Granular access tokens allow you to restrict access provided to the token based 
 - Grant tokens access to specific organizations
 - Set a token expiration date
 - Limit token access based on IP address ranges
-- Select between **read-only** or **read and write** access. For packages and scopes, **read and write** access can either **publish and stage** new versions or be **stage only** (see "[About stage-only tokens](#about-stage-only-tokens)")
+- Select between **Read-only** or **Read and write** access. For packages and scopes, **Read and write** access can either **publish and stage** new versions or be **stage only** (see "[About stage-only tokens](#about-stage-only-tokens)").
 - Configure a token to **Bypass 2FA** requirements
 
 You can create up to 1000 granular access tokens on your npm account. You can set how long your token is valid for, at least one day in the future. Each token can access up to 50 organizations, and up to either 50 packages, 50 scopes, or a combination of 50 packages and scopes. Access tokens are tied to users’ permission; hence it cannot have more permission than the user at any point in time. If a user has their access revoked from a package or an org., their granular access token will also have its access revoked from those packages or org.
@@ -51,7 +51,7 @@ When a granular access token has **Read and write** access to packages and scope
 - **Read and write (publish and stage)**: the token can publish new package versions directly to the registry.
 - **Read and write (stage only)**: the token _cannot_ publish new package versions directly. Instead, it stages a version for a package maintainer to review and promote.
 
-Stage-only tokens are a safer on-ramp for automation such as continuous integration and deployment (CI/CD). Because promoting a staged version always requires a maintainer with two-factor authentication (2FA), a leaked or misused stage-only token cannot push a new version live on its own.
+Stage-only tokens are a safer on-ramp for automation such as continuous integration and deployment (CI/CD). Because a stage-only token can't publish a new version directly — each staged version must be reviewed and promoted by a maintainer with two-factor authentication (2FA) — a leaked or misused token can't get a new version onto the registry without that maintainer approval. This narrows the risk of direct publishing; it does not otherwise restrict the token's write capabilities.
 
 A stage-only token **can**:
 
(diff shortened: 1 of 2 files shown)


content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
```diff
@@ -44,6 +44,44 @@ When you give a token access to an organization, the token can only be used for
...
+- **Read and write (stage only)**: the token _cannot_ publish new package versions directly. Instead, it stages a version for a package maintainer to review and promote.
+
+Stage-only tokens are a safer on-ramp for automation such as continuous integration and deployment (CI/CD). Because promoting a staged version always requires a maintainer with two-factor authentication (2FA), a leaked or misused stage-only token cannot push a new version live on its own.
+
+A stage-only token **can**:
+
+- Stage a new package version for review, using `npm stage publish`
+- Deprecate package versions
+- Move dist-tags
+- Unpublish package versions
```
dhei:
```suggestion
```
It can't unpublish.

content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
```diff
@@ -44,6 +44,44 @@ When you give a token access to an organization, the token can only be used for
...
+2. **Review the stage.** A maintainer can list pending stages with `npm stage list` and inspect one with `npm stage view <stage-id>` or `npm stage download <stage-id>`.
+3. **Promote the version.** A maintainer with 2FA enabled promotes the staged version to the registry with `npm stage approve <stage-id> --otp <code>`, or discards it with `npm stage reject <stage-id>`.
+
+#### Direct publishing is being deprecated
+
+Direct-publish tokens are being deprecated. The ability to publish new package versions directly with a granular access token will be removed in **January 2027**. To prepare, switch automation that publishes new versions to a **Read and write (stage only)** token together with the staging workflow described above. See also "[Account-identity actions require an interactive 2FA challenge](#account-identity-actions-require-an-interactive-2fa-challenge)" for related changes to bypass-2FA tokens.
```
dhei:
```suggestion
Bypass-2fa token with direct-publish access are being deprecated. The ability to publish new package versions directly with a granular access token will be removed in **January 2027**. To prepare, switch automation that publishes new versions to a **Read and write (stage only)** token together with the staging workflow described above. See also "[Account-identity actions require an interactive 2FA challenge](#account-identity-actions-require-an-interactive-2fa-challenge)" for related changes to bypass-2FA tokens.
```

dhei (approved):

content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
Tayvon: Thanks @dhei — I want to double-check this one before dropping the bullet, because the enforcement code and the live UI copy both say stage-only tokens *can* unpublish:

- **validate-and-store** `lib/views/publish.js` gates only on `newVersionsPublished.length > 0`; the inline comment states the enforcement is *"Scoped to new-version adds, so deprecate / unpublish / dist-tag / owner changes pass through untouched."*
- **npm-auth-ws** `authorizer.js` only blocks brand-new **packages** (`MESSAGE_403_STAGE_ONLY_NEW_PKG`); registry-frontdoor just sets the claim and enforces nothing on unpublish.
- The shipped wubwub creation UI helper text (`STAGE_ONLY_HELPER_TEXT`) also says the token *"can still deprecate versions, move dist-tags, and unpublish."*

So as currently implemented, unpublish isn't blocked. Is there a separate/planned restriction I'm missing, or should this bullet stay? Happy to remove it (and flag the UI copy as needing a fix) if unpublish is meant to be blocked — just want the docs to match the actual behavior. Leaving the bullet in for now.

Tayvon:
Addressed the review feedback in 20cf914:

- **Jan 2027 scope** (@dhei): reworded to *"Bypass-2FA tokens with direct-publish access are being deprecated…"* — matches github/npm#15610. ✅
- **Overstating security** (Copilot): rephrased to state the narrow guarantee — no direct new-version publish without maintainer approval — rather than implying broad safety. ✅
- **Permission label casing** (Copilot): title-cased **Read-only** / **Read and write** and added the trailing period. ✅

Left open for your call, @dhei: the **"Unpublish package versions"** capability bullet. Per the enforcement code (`validate-and-store` `publish.js` scopes the block to new-version adds only) and the live creation-UI helper text, stage-only tokens *can* currently unpublish — details in the thread above. Happy to drop it if unpublish is meant to be blocked.

content/integrations/integrating-npm-with-external-services/about-access-tokens.mdx
Tayvon: Confirmed and applied in 39b72d9 — removed the unpublish bullet. You're right: I traced it to the **bypass-2FA GAT allowlist** (`npm-auth-ws` `checkBypass2faGatAccess`, flag #15226), which permits publish / stage publish·approve·reject / dist-tags but **not** unpublish. `deprecate` is a `PUT` packument (the publish pathName) so it passes — exactly matching your test. Since stage-only automation tokens are the bypass-2FA population, unpublish is effectively blocked for them, so the bullet was misleading. Kept deprecate + dist-tags.

Side note: the wubwub creation-UI helper text (`STAGE_ONLY_HELPER_TEXT`) still says these tokens "can … unpublish" — likely worth a follow-up fix there too.

